PR #7, opened 2020-01-09
[TUBEMQ-14]: Some compilation errors

kaynewu:
Fixes Bug #<[TUBEMQ-14](https://issues.apache.org/jira/browse/TUBEMQ-14)>

### Motivation
Bug fix, Syntax correct

### Modifications
delete error charset


chenjunjiedada:
Can we apply plugin to strip trailing space like this? How about https://github.com/diffplug/spotless. @gosonzhang, What do you think?

kaynewu:
+1 for  https://github.com/diffplug/spotless

PR #12, opened 2020-01-15
[TUBEMQ-21]Change version number from x.y-SNAPSHOT to x.y.z-incubating-SNAPSHOT

gosonzhang:
1. Adjust version format
2. Add mailist, scm, license, issueManagement in pom.xml


lizhiboo:
+1

PR #17, opened 2020-01-19
[TUBEMQ-28] fix ignore path error

aloyszhang:
Fix path error in gitignore.
From `tube-server/test/` to `tubemq-server/src/test/`


aloyszhang:
@gosonzhang

PR #19, opened 2020-02-10
[TUBEMQ-12]Change to use Apache License V2

gosonzhang:


yiheng:
can you highlight where do we enable RAT check?

gosonzhang:
This PR only changed LICENSE of files, the " enable RAT check" is copied from Jira name

yiheng:
I suggest split the JIRA into small sub-tasks

yiheng:
IMHO, the change is not reviewable without check in the enable RAT check change. No one can review such changes. Instead, if we enable the RAT check. The change can be verified automatically by the travis build job. Then we just need to review the RAT enable check change.

guangxuCheng:
Agree with what @yiheng said.

gosonzhang:
I checked the pom.xml, the rat check already existed, but not enabled in the CI script, but after the function was enabled, the entire compilation time took more than 20 minutes and it is not over yet. 

It is better to enable RAT check locally before submitting pr or before incorporating the code, NOT in CI script.

gosonzhang:
After enable the RAT check, I found that the PR compilation time has been extended a lot, so I reverted the modification of the travis-ci script.

Everyone can enable CAT check by adding "apache-rat: check" to the mvn command line

PR #21, opened 2020-02-11
[TUBEMQ-29] Change the package name to "org.apache.tubemq."

gosonzhang:


chenjunjiedada:
Seems it is a huge change, I would suggest merging once CI passed. @gosonzhang what do you think?

gosonzhang:
I'm adding more test cases, need wait some time and then merge them.

A large-scale change is inevitable if we need change the package name.

PR #23, opened 2020-02-16
[TUBEMQ-32] rename file path in tubemq-client module to match package name

aloyszhang:
This PR renames file path in tubemq-client module to match package name


aloyszhang:
cc @gosonzhang

gosonzhang:
Thanks @aloyszhang ,

PR #24, opened 2020-02-18
[TUBEMQ-33] refactor enum implemention of inner annoynmous

aloyszhang:
refactor enum implemention from inner annoynmous to constructor with necessary parameters


aloyszhang:
cc @gosonzhang

PR #44, opened 2020-04-05
[TUBEMQ-57] fix some typos & todo

Technoboy-:


tubemq-server/src/main/java/org/apache/tubemq/server/master/web/action/screen/Webapi.java
```diff
@@ -191,8 +191,8 @@ private StringBuilder processModifyOperate(HttpServletRequest req, final String
         BrokerConfManage brokerConfManage = this.master.getMasterTopicManage();
         if ("admin_add_broker_configure".equals(method)) {
             strBuffer = webBrokerDefConfHandler.adminAddBrokerDefConfEntityInfo(req);
-        } else if ("admin_bath_add_broker_configure".equals(method)) {
```
gosonzhang, line 194: First of all, thanks @Technoboy- for finding so many problems!

Then, because the token string of the method has been released to the public, the modifications in webapi.java are correct, but they will lead to incompatibility, so, these parts can't be modified.

gosonzhang: You can see that "admin_bath_add_broker_configure"，"admin_bath_add_new_topic_record", etc. are fixed tokens, which is passed during HTTP access

Technoboy-: ok

PR #45, opened 2020-04-07
[TUBEMQ-58] fix some typos

Technoboy-:


tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/DataConverterUtil.java
```diff
@@ -108,17 +108,12 @@
...
-                        String[] strTopicInfo =
-                                strTopicInfoSet[i].split(TokenConstants.ATTR_SEP);
-                        BrokerInfo brokerInfo =
-                                brokerInfoMap.get(Integer.parseInt(strTopicInfo[0]));
-                        if (brokerInfo != null) {
-                            topicList.add(new TopicInfo(brokerInfo,
-                                    strInfo[0], Integer.parseInt(strTopicInfo[1]),
-                                    Integer.parseInt(strTopicInfo[2]), true, true));
-                        }
+                    String[] strTopicInfo = strInfo[1].split(TokenConstants.ATTR_SEP);
```
gosonzhang, line 111: There is a problem in this place, the info is actually an array, the content is similar to "topicName TokenConstants.SEGMENT_SEP [[TokenConstants.ARRAY_SEP] brokerId TokenConstants.ATTR_SEP partitionNum TokenConstants.ATTR_SEP topicStoreNum]",  if modified in this way,, it will cause the loss of topic metadata

Technoboy-: yes, it is.
last night, i debug the the heartbeat response, and check the UT of DataConverterUtil#convertTopicInfo, it's green.  but not checking the master part of the heartbeat request (TMaster#getProducerTopicPartitionInfo).

PR #49, opened 2020-04-11
[TUBEMQ-63] Replace force local dependency of protoc with auto-downloading

TisonKun:
Replace `com.github.igor-petruk.protobuf: protobuf-maven-plugin` with `org.xolstice.maven.plugins: protobuf-maven-plugin` so that we don't have to pre-configure a local protoc but rely on auto-downloading.


gosonzhang:
Thanks for TisonKun, 

TubeMQ uses Protocol Buffers, mainly for borrowing and only borrowing pb's ZigZag, Varints encoding scheme to facilitate the construction of the underlying network message package. It doesn't matter which version of pb is specific, so you can see that this version is very old Old.

Actually, we found that if this piece is bound to pb, the compile time requires the business to install this component in advance, and at the same time, the issued sdk is likely to form a dependency conflict with the business's pb version.

Considering that TubeMQ is not a component with a wide audience such as Hbase, whether it is a good method to install the script compilation embedded dependency package, I don't know， and I worry that the first-time user may be confused and increase the barrier to use. If you need this when you use it, you can replace it directly. 

Subsequent considerations:  deleting PB component dependencies may be a better choice.

netroby:
@gosonzhang agree to delete protobuf, it not the must requirement .

guangxuCheng:
@TisonKun Can you fix these conflicts? This is a good idea. If no objection, i will merge this PR

netroby:
anyway, this pull request solve one problem , User may not want download protoc, Fuck protobuf as my side(I hate protobuf)

netroby:
For the bad side. Import the third party dependencies not always good things. especially for opensource project.

netroby:
If any one, such as a hacker inject backdoor with that  compile plugin. may dangerous . It is possible lost control.

TisonKun:
> @TisonKun Can you fix these conflicts? This is a good idea. If no objection, i will merge this PR

Done. Except the document side. Could you guide me to the new place of our documents?

guangxuCheng:
> > @TisonKun Can you fix these conflicts? This is a good idea. If no objection, i will merge this PR
> 
> Done. Except the document side. Could you guide me to the new place of our documents?

Hi @TisonKun documents  had been moved  to https://github.com/apache/incubator-tubemq-website/tree/master/docs

TisonKun:
@guangxuCheng thanks! Merging this one and will update document.

TisonKun:
FYI:

https://github.com/apache/incubator-tubemq-website/commit/c376aac05b01f160043119782b1729dcf05b9311
https://github.com/apache/incubator-tubemq-website/commit/b0527875232032829e7473d085d1e997e44a55cf

PR #52, opened 2020-04-13
[TUBEMQ-66] TubeSingleSessionFactory shutdown bug

Technoboy-:


Technoboy-:
> After this modification, there will be problems in Flink-like usage scenarios:
> 
> after the reference count is changed to the boolean value, if the TubeSingleSessionFactory object is initialized multiple times in the same process, as long as there is a shutdown() call, then the other executing logics in the same process will throw wrong infos.

So we have to delete the referenceCounter.incrementAndGet() in the constructor?

gosonzhang:
The TubeMultiSessionFactory class has implemented your approach. The TubeSingleSessionFactory class is mainly used in the scheme of requiring shared connections

Technoboy-:
> The TubeMultiSessionFactory class has implemented your approach. The TubeSingleSessionFactory class is mainly used in the scheme of requiring shared connections



> The TubeMultiSessionFactory class has implemented your approach. The TubeSingleSessionFactory class is mainly used in the scheme of requiring shared connections

yes, i see. But not clear why call another time for referenceCounter.incrementAndGet()

PR #56, opened 2020-04-17
[TUBEMQ-73] remove duplicate codes & some minor updates

Technoboy-:


tubemq-server/src/main/java/org/apache/tubemq/server/broker/BrokerServiceServer.java
```diff
@@ -673,7 +673,7 @@ public SendMessageResponseB2P sendMessageP2B(SendMessageRequestP2B request,
             final MessageStore store =
                     this.storeManager.getOrCreateMessageStore(reqTopic, partition);
             final long messageId = this.idWorker.nextId();
-            if (store.appendMsg(messageId, dataLength, checkSum, msgData,
+            if (store.appendMsg(messageId, dataLength, request.getCheckSum(), msgData,
```
gosonzhang, line 676: This place needs to be reserved. Even if the client does not do CRC, it needs to check whether it is consistent with the storage during consumption. After removal, this logic will be lost

Technoboy-: yes, I see.

Technoboy-: I will repost laster

PR #61, opened 2020-04-21
[TUBEMQ-76] rename the classes

Technoboy-:


tubemq-server/src/test/java/org/apache/tubemq/server/broker/metadata/BrokerMetadataManageTest.java
```diff
@@ -1,74 +0,0 @@
-/**
```
aloyszhang, line 1: Why this file is deleted?

Technoboy-: I will check it right now , seems omit

Technoboy-: It's .gitignore that contains 'tubemq-server/src/test/',  omit BrokerMetadataManageTest.java

PR #64, opened 2020-04-23
[TUBEMQ-81] Fix LogClearRunner bug && format BrokerDefMetadata

Technoboy-:


tubemq-server/src/main/java/org/apache/tubemq/server/broker/msgstore/MessageStoreManager.java
```diff
@@ -703,7 +703,7 @@ public void run() {
                         .append(tubeConfig.getLogClearupDurationMs()).toString());
                 sBuilder.delete(0, sBuilder.length());
             }
-            if (!expiredTopic.isEmpty()) {
+            if (expiredTopic.isEmpty()) {
```
aloyszhang, line 706: Print log here indicates that some expired file has been deleted, so I think we cann't make this change .

aloyszhang: Maybe move this log print to  [the end of the loop of file delete](https://github.com/apache/incubator-tubemq/blob/master/tubemq-server/src/main/java/org/apache/tubemq/server/broker/msgstore/MessageStoreManager.java#L697)

aloyszhang: Mabye move this log print to  [the end of file delete loop](https://github.com/apache/incubator-tubemq/blob/master/tubemq-server/src/main/java/org/apache/tubemq/server/broker/msgstore/MessageStoreManager.java#L697) is better.

gosonzhang: I agree with @aloyszhang ‘s view. If this place is judged to be empty, not only will the previous end-of-print mark for deleting the file log printing be lost, but also the useless printing log will increase(most of the time, the expired file check returns empty).

Technoboy-: Yes, you are right.  I made a mistake of the end of the If logic. I will repost a pull request

PR #82, opened 2020-05-15
[TUBEMQ-121] Modify the maven jdk compilation version

viviel:
Modify the maven jdk compilation version to solve the syntax error during development and compilation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -100,6 +100,13 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
+                    <!--
+                      Note:
+                      The client and core modules use jdk1.7,
+                      and the server module uses 1.8.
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -100,8 +100,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>1.7</source>
```
gosonzhang, line 103: I think it would be perfect if you add a comment here to tell the following partners why this setting is necessary: the JAVA version of the environment where the client is located may be different from the server, tubemq-client, tubemq-core should maintain the 1.7 version.

gosonzhang: And this pr need a jira number ，like [TubeMQ-xxx] xxxxx

viviel: Okay, I will add a comment here

PR #84, opened 2020-05-15
[TUBEMQ-121] Modify the maven jdk compilation version

viviel:
1. Modify the maven jdk compilation version to solve the syntax error during development and compilation
2. Add maven jdk compilation version comment


gosonzhang:
Thanks @viviel

PR #93, opened 2020-05-22
[TUBEMQ-154] Modify the wrong comment & Minor changes for example module

viviel:
1. Modify the wrong comment
```
    /**
     * Send message synchronous. More efficient and recommended.
     */
    public void sendMessageAsync() {}
```
change to
```
    /**
     * Send message asynchronous. More efficient and recommended.
     */
    public void sendMessageAsync() {}
```
2. Optimize variable naming


gosonzhang:
Thanks @viviel !

We need to check carefully when uploading the code.

PR #97, opened 2020-05-25
[TUBEMQ-141] Remove the requirement to provide localHostIP

viviel:
1. Remove the requirement to provide localHostIP
2. Optimize the TubeClientConfig constructor to remove some explicit exceptions

### DIFF
--- a/tubemq-client/src/main/java/org/apache/tubemq/client/config/TubeClientConfig.java
+++ b/tubemq-client/src/main/java/org/apache/tubemq/client/config/TubeClientConfig.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
@@ -92,23 +92,26 @@ public class TubeClientConfig {
     // TLS configuration.
     private TLSConfig tlsConfig = new TLSConfig();
 
-
-    public TubeClientConfig(final String localHostIP, final String masterAddrInfo) throws Exception {
-        if (TStringUtils.isBlank(localHostIP)) {
-            throw new IllegalArgumentException("Illegal parameter: localHostIP is blank!");
-        }
-        // Not allow to set local host ip to 127.0.0.1.
-        if ("127.0.0.1".equals(localHostIP)) {
-            throw new IllegalArgumentException("Illegal parameter: localHostIP can't set to 127.0.0.1");
-        }
-        if (TStringUtils.isBlank(masterAddrInfo)) {
-            throw new IllegalArgumentException("Illegal parameter: masterAddrInfo is Blank!");
-        }
-        this.masterInfo = new MasterInfo(masterAddrInfo);
-        AddressUtils.validLocalIp(localHostIP.trim());
+    public TubeClientConfig(String masterAddrInfo) {
+        this(new MasterInfo(masterAddrInfo));
     }
 
-    public TubeClientConfig(final String localHostIP, final MasterInfo masterInfo) throws Exception {
+    public TubeClientConfig(MasterInfo masterInfo) {
+        if (masterInfo == null) {
+            throw new IllegalArgumentException("Illegal parameter: masterAddrInfo is null!");
+        }
+        this.masterInfo = masterInfo.clone();
+        String iPv4LocalAddress = AddressUtils.getIPV4LocalAddress();
+        AddressUtils.setLocalAddress(iPv4LocalAddress);
+    }
+
+    @Deprecated
+    public TubeClientConfig(final String localHostIP, final String masterAddrInfo) {
+        this(localHostIP, new MasterInfo(masterAddrInfo));
+    }
+
+    @Deprecated
+    public TubeClientConfig(final String localHostIP, final MasterInfo masterInfo) {
         if (TStringUtils.isBlank(localHostIP)) {
             throw new IllegalArgumentException("Illegal parameter: localHostIP is blank!");
         }
--- a/tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
+++ b/tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one or more
  * contributor license agreements.  See the NOTICE file distributed with
  * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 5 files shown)


tubemq-client/src/main/java/org/apache/tubemq/client/config/TubeClientConfig.java
```diff
@@ -92,23 +92,19 @@
...
-        if (TStringUtils.isBlank(masterAddrInfo)) {
-            throw new IllegalArgumentException("Illegal parameter: masterAddrInfo is Blank!");
-        }
-        this.masterInfo = new MasterInfo(masterAddrInfo);
-        AddressUtils.validLocalIp(localHostIP.trim());
+    public TubeClientConfig(MasterInfo masterInfo) {
+        this(AddressUtils.getLocalIPv4Address(), masterInfo);
+    }
+
+    public TubeClientConfig(final String localHostIP, final String masterAddrInfo) {
```
gosonzhang, line 103: need to add a @Deprecated tag on this function

tubemq-client/src/main/java/org/apache/tubemq/client/config/TubeClientConfig.java
```diff
@@ -92,23 +92,19 @@
...
+        this(localHostIP, new MasterInfo(masterAddrInfo));
     }
 
-    public TubeClientConfig(final String localHostIP, final MasterInfo masterInfo) throws Exception {
+    public TubeClientConfig(final String localHostIP, final MasterInfo masterInfo) {
```
gosonzhang, line 107: need to add a @Deprecated tag on this function

viviel: If it is confirmed that the IP will be automatically and uniformly configured later, it is actually better. But will some people specify the IP by themselves or fill in the IP by specifying the network interface?This method is provided here:
```
AddressUtils.getIPV4AddressByNetworkInterfaceName("eth0");
```
I am not sure if there is such a scene

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -130,4 +134,100 @@ public static String getRemoteAddressIP(Channel channel) {
         }
         return strRemoteIP;
     }
+
```
gosonzhang, line 137: What if there is only IPV6 on the machine?

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -130,4 +134,100 @@ public static String getRemoteAddressIP(Channel channel) {
...
+            try {
+                return InetAddress.getLocalHost().getHostAddress();
+            } catch (UnknownHostException e) {
+                String errorMsg = "Unable to obtain local IP, please manually specify the local IP address";
+                throw new AddressException(errorMsg, e.getCause());
+            }
+        }
+
+        for (String ip : ips) {
+            if (ip.startsWith("192.168")) {
```
gosonzhang, line 151: Why exclude "192.168" separately?

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -130,4 +134,100 @@ public static String getRemoteAddressIP(Channel channel) {
...
+                continue;
+            }
+
+            return ip;
+        }
+
+        return ips.get(ips.size() - 1);
```
gosonzhang, line 158: If ips is empty, how to do?

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -34,31 +40,29 @@ public static synchronized String getLocalAddress() throws Exception {
...
+
+        Enumeration<NetworkInterface> allInterface = listNetworkInterface();
+
         return checkValidIp(allInterface, currLocalHost);
     }
 
-    private static boolean checkValidIp(Enumeration<NetworkInterface> allInterface,
-                                        String currLocalHost) throws Exception {
-        String localIp = null;
+    private static boolean checkValidIp(Enumeration<NetworkInterface> allInterface, String currLocalHost) {
```
gosonzhang, line 54: It is possible that the user prefers to obtain IP addresses from a certain network card first, and then the IPV4 type address is the best choice, so can the function achieve such an effect: 
1. the program first obtains the IP address from the specified network card, and prefers the IPV4 address; 
2. the specified network card does not Get IP from other network card when IPV4 address; 
3. if there are no IPV4 address, I choose IPV6; 
4. when all network cards can not get IP address, then get from InetAddress object?

viviel: The initial idea is only for IPV4, if the IPV4 address cannot be obtained on the machine. Will cause an exception to the user, let the user specify the IP address, which is why the two constructors are not marked with @deprecated annotation.

Now, I understand that IPV4 and IPV6 are used together. In that case, I will reorganize the process

There are two options:

First, get the address completely automatically
1. Priority access to IPV4 address.
2. If not available, obtain an IPV6 address.
3. No IP can be obtained, choose one of the two below
   1) Throw the exception to the user
   2) Get IP from InetAddress

Second, give priority to obtain IP from the designated network interface
1. Try to obtain IPV4 from the specified network card. If not, obtain IPV4 from other network interfaces.
2. If IPV4 is not obtained in the previous step, try to obtain IPV6 from the specified network interface. Otherwise, obtain IPV6 from other network interfaces.
3. No IP can be obtained, choose one of the two below
   1) Throw the exception to the user
   2) Get IP from InetAddress

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -34,31 +39,33 @@ public static synchronized String getLocalAddress() throws Exception {
         return localIPAddress;
     }
 
-    public static boolean validLocalIp(String currLocalHost) throws Exception {
+    public static synchronized void setLocalAddress(String localIPAddress) {
+        AddressUtils.localIPAddress = localIPAddress;
+    }
+
+    public static boolean validLocalIp(String currLocalHost) {
```
gosonzhang, line 46: The validLocalIp function seems to be superfluous and should be removed along with the sub-functions

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
gosonzhang: ok！

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/AddressUtils.java
```diff
@@ -130,4 +137,50 @@ public static String getRemoteAddressIP(Channel channel) {
...
+            throw new AddressException(errorMsg, e);
+        }
+    }
+
+    private static String doGetIPV4LocalAddress() throws SocketException, UnknownHostException {
+        Enumeration<NetworkInterface> enumeration = NetworkInterface.getNetworkInterfaces();
+
+        while (enumeration.hasMoreElements()) {
+            NetworkInterface networkInterface = enumeration.nextElement();
+            if ("docker0".equals(networkInterface.getName()) || !networkInterface.isUp()) {
```
gosonzhang, line 163: Need to add NetworkInterface isLoopback() status judgment

PR #99, opened 2020-05-26
[TUBEMQ-159] Fix some typos

cku328:
I found some typos in TubeMQ code and fixed it.

### DIFF
--- a/docs/http_access_API_definition.md
+++ b/docs/http_access_API_definition.md
@@ -740,7 +740,7 @@ __Request__
 |---|---|---|---|
 |groupName|yes|the group name to set flow control rule|String|
 |enableBind|no|whether to bind consuming permission, default value 0 means disable|int|
-|allowBClientRatio|no|the ratio of the number of the consuming target's broker against the number of client in consuming group|int|
+|allowedBClientRate|no|the ratio of the number of the consuming target's broker against the number of client in consuming group|int|
 |createUser|yes|the creator|String|
 |createDate|yes|the creating date in format `yyyyMMddHHmmss`|String|
 |confModAuthToken|yes|the authorized key for configuration update|String|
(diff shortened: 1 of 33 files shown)


docs/http_access_API_definition.md
```diff
@@ -740,7 +740,7 @@ __Request__
 |---|---|---|---|
 |groupName|yes|the group name to set flow control rule|String|
 |enableBind|no|whether to bind consuming permission, default value 0 means disable|int|
-|allowBCleintRatio|no|the ratio of the number of the consuming target's broker against the number of client in consuming group|int|
+|allowBClientRatio|no|the ratio of the number of the consuming target's broker against the number of client in consuming group|int|
```
gosonzhang, line 743: This needs to be modified together with the external api parameters and configuration parameters, or will the document be inconsistent with the actual field

But after this modification, it will affect the already used business adjustments. How can it be compatible?

cku328:
Thanks @gosonzhang for the review and comments.

I also agree that modifying the external API parameter field is risky, and I re-checked my commit and how it relates to TubeMQ code.

I compared the differences between Chinese and English versions of the HTTP API Definition Guide (`docs/http_access_API_definition.md` and `docs/http_access_API_definition_cn.xls`).

From the Chinese version, I found that the parameter field should be "**allowedBClientRate**", and I verified from some code that this name is correct, like [this](https://github.com/apache/incubator-tubemq/blob/02220e747f4251fa4ddf0eef305bfad31648428e/tubemq-server/src/main/java/org/apache/tubemq/server/master/web/handler/WebAdminGroupCtrlHandler.java#L1237:L1238).
image

And the parameter field "**modfiyUser**", I also confirmed that it's just a typo in the document.

I updated my patch based on the above description.

PR #104, opened 2020-05-29
[TUBEMQ-104] Substitute Chinese comments with English - core module

viviel:
Replace the Chinese comment of the core module with English


aloyszhang:
+1

PR #118, opened 2020-06-02
protobuf version adapt

yayoec:
OS: Mac os X 10.15.3
Java: 14.0.1
maven: 3.6.3
libprotoc: 3.12.2

this env will cause compiling error

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
         <exec.maven.version>1.6.0</exec.maven.version>
         <build.helper.maven.version>3.0.0</build.helper.maven.version>
         <checkstyle.version>6.19</checkstyle.version>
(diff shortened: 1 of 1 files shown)


guangxuCheng:
hi @yayoec, Thanks for your contributing.

TubeMQ use jira for issue management. Before submitting a PR, you need to create a new issue on jira. see:[how-to-contribute](http://tubemq.apache.org/zh-cn/docs/development/how-to-contribute.html)

By the way，you need not to pre-configure a local protoc since [TUBEMQ-63](https://issues.apache.org/jira/browse/TUBEMQ-63), protoc will be downloaded automatically during build source.

As you said, the protobuf version is too old, we had created a issue to solve it, see: [TUBEMQ-203](https://issues.apache.org/jira/browse/TUBEMQ-203)

PR #120, opened 2020-06-02
[TUBEMQ-17]support standalone mode for development and experience

dockerzhang:
support standalone mode for development and experience


kaynewu:
> support standalone mode for development and experience

 hi， @dockerzhang  dockerzhang if windows development environment can support standalone mode it will be more helpful.

netroby:
docker for windows may working not as well as linux/mac. It's hard for explain how to let it working perfect. 

User may using different Windows. Windows 7, windows 8, Windows 10

gosonzhang:
@kaynewu @netroby, Let's discuss this matter on jira,  what should we do on windows is more appropriate? 

The discussion on jira is more and more inherited

dockerzhang:
the best way is to install [Docker for Windows](https://docs.docker.com/docker-for-windows/), and i have tested the image on windows 10, it's works fine.

PR #121, opened 2020-06-02
[TUBEMQ-172] simplify start/stop script

cku328:
I merged the features of the 2 scripts (master.sh/broker.sh), which support commands to start, stop and restart on either the master or broker server.

When the directory path is added to the path environment variable, it works like this:
image

I wonder if this solves the problem with this jira?

### DIFF
--- a/bin/tubemq
+++ b/bin/tubemq
@@ -1,0 +1,163 @@
+#!/bin/bash
+
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


dockerzhang:
@cku328  good job.  Could you please fix the [user guide](https://github.com/apache/incubator-tubemq/blob/master/docs/tubemq_user_guide.md)  about start/stop at same time?

cku328:
Thanks @dockerzhang  for the review.
I have updated the user guide.

gosonzhang:
Thanks @cku328

cku328:
Thanks @gosonzhang for reviewing and merging this.

PR #125, opened 2020-06-04
[TUBEMQ-185] Improve producer implementation - support data compression

Technoboy-:


tubemq-client/src/main/java/org/apache/tubemq/client/producer/SimpleMessageProducer.java
```diff
@@ -335,11 +336,13 @@ private void checkMessageAndStatus(final Message message) throws TubeClientExcep
     }
 
     private byte[] encodePayload(final Message message) {
```
gosonzhang, line 338: It would be better to put the message compression operation in the Message class. It is open to the user and can avoid the situation where the Message data has been compressed when it is sent and then compressed again.

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/CompressionType.java
```diff
@@ -0,0 +1,94 @@
+/**
```
gosonzhang, line 1: Compression processing and compression enumeration need to be separated: at present, there is no problem when there is only one compression method, but subsequent compression methods are added, some methods need to be initialized, and some require other processing, the enumeration class will become unstable, and need to be adjusted frequently. 

So enumerating compression methods and specific compression classification will be better to solve this problem

tubemq-core/src/main/java/org/apache/tubemq/corebase/utils/DataConverterUtil.java
```diff
@@ -233,8 +233,17 @@
                     payloadDataLen -= attrLen;
                 }
             }
-            final byte[] payload = new byte[payloadDataLen];
+            byte[] payload = new byte[payloadDataLen];
             System.arraycopy(payloadData.array(), readPos, payload, 0, payloadDataLen);
+            String compressionType = "";
```
gosonzhang, line 238: The extraction of the compression type value needs to be performed by calling the decoding function of the Message attribute. With the iteration of the version, it may be changed later, and it is more appropriate to process it by the method

tubemq-client/src/main/java/org/apache/tubemq/client/producer/SimpleMessageProducer.java
```diff
@@ -335,11 +336,13 @@ private void checkMessageAndStatus(final Message message) throws TubeClientExcep
...
-        final byte[] payload = message.getData();
-        final String attribute = message.getAttribute();
+        byte[] payload = message.getData();
+        String attribute = message.getAttribute();
         if (TStringUtils.isBlank(attribute)) {
-            return payload;
+            attribute = "";
         }
+        attribute = attribute + "," + TokenConstants.TOKEN_COMPRESS_TYPE + producerConfig.getCompressionType().name();
```
gosonzhang, line 344: The attribute item in the attribute needs to be added in a=b, c=d mode and has a fixed format. It is best to call the method of Message to add the attribute item in this place, instead of seeming to manually stitch together to avoid incompatibility after adjusting the encoding

tubemq-client/src/main/java/org/apache/tubemq/client/producer/SimpleMessageProducer.java
```diff
@@ -321,7 +322,7 @@ private void checkMessageAndStatus(final Message message) throws TubeClientExcep
         builder.setTopicName(partition.getTopic());
         builder.setPartitionId(partition.getPartitionId());
         builder.setData(ByteString.copyFrom(encodePayload(message)));
-        builder.setFlag(MessageFlagUtils.getFlag(message));
+        builder.setFlag(MessageFlagUtils.getFlag(message, true));
```
gosonzhang, line 325: The function of payload data process is recommended to be placed on line 321, and the attribute need to be adjusted by the function, 
 and there is no need to add a mandatory flag setting, which will make the maintenance personnel confused later.

PR #130, opened 2020-06-04
[TUBEMQ-206] Delete residual files after executed unit tests

duli559:
There are some residual files after executed unit tests by 'tubemq-server' module, create temporary files to avoid this problem.

Jira link with: https://issues.apache.org/jira/browse/TUBEMQ-206


aloyszhang:
Nice contribution. LGTM

PR #131, opened 2020-06-04
[TUBEMQ-203] Upgrade Protocol Buffer component version

viviel:
Remove the steps that need to explicitly configure protocol buffer on the local machine, and directly use Maven plugin to complete, simplifying the process of contributor's configuration of Pb environment

### DIFF
--- a/tubemq-core/pom.xml
+++ b/tubemq-core/pom.xml
@@ -31,6 +31,13 @@
     <description>Core functionality for TubeMQ</description>
 
     <build>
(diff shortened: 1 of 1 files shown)


guangxuCheng:
duplicate of #49 ?

TisonKun:
Closed by #49 . Thanks for nudging this work.

TisonKun:
@viviel from the original ticket it is still valid if you try to shade protobuf deps for goods.

PR #133, opened 2020-06-05
[TUBEMQ-214] upgrade je.version to 18.3.12

netroby:


netroby:
https://search.maven.org/artifact/com.sleepycat/je

Version 7.3.7 missing from public MVN repository.
So shall we upgrade to public release version?

The only version 7.3.7 can be found on oracle.com . 

the only solution is add mvn repository as depencency

```
<repositories>
     <repository>

         <id>oracleReleases</id>
         <name>Oracle Released Java Packages</name>
         <url>http://download.oracle.com/maven</url>
         <layout>default</layout>

     </repository>
</repositories>
```

May we really want to add a company selfhost maven repository as dependency?

gosonzhang:
The download path is changed, I'll fixed!

image

netroby:
As an opensource project. we should carefully handle third party dependency. 
Should not depends on company self-hosted Maven Repository.
There is no guarated safe and stable.

gosonzhang:
I submitted this issue to resolve this question: https://issues.apache.org/jira/projects/TUBEMQ/issues/TUBEMQ-217?filter=allopenissues

gosonzhang:
Yes, is there any good way?

At the beginning, the system requires users to download these third-party components, such as proto buffer and je, to allow users to download directly. But for more convenient use, an automatic download link is added. So if this content is adjusted by a third party, then the compilation there will be problems; but I think this is not the reason to upgrade the third-party library version without verification.

netroby:
Replace 7.3.7 with a valid download repository is an tempory  resolution. 
For long term maintain prespective. upgrade it to public maven repository may be the only way .

gosonzhang:
I agree with you, but if the new version is directly introduced into the system as a white mouse without verification, it is likely to affect the external services provided by the project, so the method we usually adopt is to select the new version if necessary and pass the verification, later release and upgrade

netroby:
I am not ask merge this PR, i am just explain it . This pull reuqest can be hold on. and wait for next time if somebody like to made some improvment .  

I 'll close it for now.

PR #134, opened 2020-06-05
[TUBEMQ-216] use ThreadUtil.sleep replace Thread.sleep

zhenjunMa:
use ThreadUtil.sleep replace Thread.sleep to avoid too many try/catch
except we need to do something by process this exception


zhenjunMa:
this pr not  accord with specifications, i will retry

PR #137, opened 2020-06-06
[TUBEMQ-222] Optimize code: Unnecessary boxing/unboxing conversion

cku328:
- Replacing valueOf() with a static method of parseXXX()
- Remove unnecessary 'String.valueOf()' call in BrokerAdminServlet#adminGetMemStoreStatisInfo

[Details on JIRA.](https://issues.apache.org/jira/projects/TUBEMQ/issues/TUBEMQ-222)


cku328:
Thanks @Technoboy- and @gosonzhang for the review.

PR #140, opened 2020-06-08
[TUBEMQ-228] Fixed: Exception used

viviel:
Fix exception use
`Exception -> IllegalArgumentException`
Example
```
if (TStringUtils.isBlank(strTimeVal)) {
    throw new Exception(strValName + " value is null or blank of "
            + ruleType + " limit rule!");
}

change to 

if (TStringUtils.isBlank(strTimeVal)) {
    throw new IllegalArgumentException(strValName + " value is null or blank of "
            + ruleType + " limit rule!");
}
```


dockerzhang:
+1

viviel:
Oh, I'm sorry, but this is specially shown to throw the inspected exception. I will close this pr

PR #143, opened 2020-06-09
[TUBEMQ-124] Structured index storage

hystericalhell:
**Initial Prototype** for demonstration only
Further details needs to be discussed. Currently it should be incomplete, and it *will* be some time.

Here the major issue directly from initial plan in TUBEMQ-124, I think, contains (but not limited to):
1. Bloom Filter, as a probability prediction set, does **not** support a **remove** operation, so we have to choose an *appropriate* time to rebuild it after a long run with massive topics in and out, which is such an expensive thing just like the GC STP (stop-the-world);
2. Fine-grained control might be introduced to scatter the impact while exploit the negative-prediction efficiency, like we could do something like Java 8 revised ConcurrentHashMap, we could setup Bloom Filter on Segment (even available for memory prediction), and apply something like URA (unbalanced resource allocation) to better concentrate messages of the same topic into the same segment.
Please review the prototype about the overall architecture of how Bloom Filter (my opinion) could behave, and let's discuss something in details about above factors, if any or more.

Details: [TUBEMQ-124](https://issues.apache.org/jira/projects/TUBEMQ/issues/TUBEMQ-124)
*The same would be synchronized to original issue comments*

### DIFF
--- a/conf/broker.ini
+++ b/conf/broker.ini
@@ -38,7 +38,8 @@ transferSize= 524288
 loadMessageStoresInParallel=true
 ; timeout of consumer heartbeat, optional; default is 30s
 consumerRegTimeoutMs=35000
-
+; enable MessageFileStore index, 0=disable, +Integer=<estimated clients amount scale>
+_mfsIndexScale=0
 
 [zookeeper]
 ; root path of TubeMQ znodes on ZK
--- a/pom.xml
+++ b/pom.xml
@@ -351,7 +351,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>13.0</version>
+                <version>23.0</version>
             </dependency>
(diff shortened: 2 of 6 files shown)


hystericalhell:
Updated:
As discussed with Guocheng Zhang, there are a few facts current system has implemented:
1. A SegmentList consists of only ONE topic, the filter is actually to distinguish clients;
2. Messages in Segments are naturally sequenced by time, which means there's only APPEND operation occurs for messages;
3. Segment would always be detached and removed when it's Expired, and it would be removed as a whole, not any part of it or a few messages in it.
Currently it's still on-going, current submission is still an intermediate solution.

hystericalhell:
Add IndexScale control in config file, while resolve estimated scale issue by accepting parameter from user.

gosonzhang:
I'll close this pr, thanks  @hystericalhell

PR #147, opened 2020-06-09
[TUBEMQ-230] upgrade test dependency, support jdk13

tensorchen:
See https://issues.apache.org/jira/projects/TUBEMQ/issues/TUBEMQ-230


gosonzhang:
+1

PR #148, opened 2020-06-10
[TUBEMQ-232] TubeBroker#register2Master, reconnect and wait

viviel:
If the broker fails to register with the master, a total of 5 reconnections will be attempted. But I think we should pause the thread for a period of time every time the connection fails to be reconnected
```
for (int i; i < reconnectionTimes; i++) {
    try {
        tryConnect();
    } catch (Throwable e) {
        try {
            TimeUnit.MILLISECONDS.sleep(200);
        } catch (InterruptedException ignored) {
            //ignore interrupted
        }
    }
}
```

### DIFF
--- a/tubemq-server/src/main/java/org/apache/tubemq/server/broker/TubeBroker.java
+++ b/tubemq-server/src/main/java/org/apache/tubemq/server/broker/TubeBroker.java
@@ -430,6 +430,7 @@ public class TubeBroker implements Stoppable {
                 if (remainingRetry == 0) {
                     throw new StartupException("Register to master failed!", e);
                 }
+                ThreadUtils.sleep(200);
             }
         }
     }


tubemq-server/src/main/java/org/apache/tubemq/server/broker/TubeBroker.java
```diff
@@ -430,6 +430,11 @@ private void register2Master() throws StartupException {
                 if (remainingRetry == 0) {
                     throw new StartupException("Register to master failed!", e);
                 }
+                try {
+                    TimeUnit.MILLISECONDS.sleep(200);
```
zhenjunMa, line 434: use ThreadUtil.sleep to replace this

viviel: Ha, thanks for the reminder

PR #157, opened 2020-06-18
[TUBEMQ-197] Support TubeMQ connector for Apache Flink

duli559:
1. Support tubemq flink SinkFunction.
2. Support Table interface for TubeMQ flink connector

Jira Link: https://issues.apache.org/jira/browse/TUBEMQ-197


tubemq-connectors/tubemq-connector-flink/src/main/java/org/apache/flink/connectors/tubemq/TubemqSinkFunction.java
```diff
@@ -0,0 +1,178 @@
...
+        int retries = 0;
+        Exception exception = null;
+
+        while (maxRetries <= 0 || retries < maxRetries) {
+
+            try {
+                byte[] body = serializationSchema.serialize(in);
+                Message message = new Message(topic, body);
+                if (StringUtils.isNotBlank(tid)) {
+                    message.setAttrKeyVal(TID, tid);
```
gosonzhang, line 137: If the message carries tid, and the business needs to filter consume according to this field, it is also necessary to call putSystemHeader(), so that the consumer can do filter consumption

duli559: ok

lamberliu:
reopen

PR #163, opened 2020-06-19
[TUBEMQ-246] remove broker ip check

dockerzhang:
remove broker IP check, the hostname with special meaning works fine.

### DIFF
--- a/tubemq-server/src/main/java/org/apache/tubemq/server/common/utils/WebParameterUtils.java
+++ b/tubemq-server/src/main/java/org/apache/tubemq/server/common/utils/WebParameterUtils.java
@@ -631,8 +631,7 @@ public class WebParameterUtils {
                 continue;
             }
             String brokerIp =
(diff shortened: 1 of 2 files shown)


tubemq-server/src/main/java/org/apache/tubemq/server/common/utils/WebParameterUtils.java
```diff
@@ -631,8 +631,7 @@ public static String checkAndGetFilterConds(String inFilterConds,
                 continue;
             }
             String brokerIp =
-                    validAddressParameter("brokerIp", strBrokerIps[i],
-                            TBaseConstants.META_MAX_BROKER_IP_LENGTH, true, "");
+                    checkParamCommonRequires("brokerIp", strBrokerIps[i], true);
```
gosonzhang, line 634: In order to better expand, it is necessary to delete the IP value check, but users need to pay attention to the accuracy of the IP value when configuring the Broker’s IP

PR #188, opened 2020-07-06
[TUBEMQ-268]Tubemq client cpp log module

charlely:
support cpp client sdk log module

### DIFF
--- a/tubemq-client-twins/tubemq-client-cpp/example/log/main.cc
+++ b/tubemq-client-twins/tubemq-client-cpp/example/log/main.cc
@@ -1,0 +1,28 @@
+
+#include <chrono>
+#include <exception>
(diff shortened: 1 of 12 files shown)


tubemq-client-twins/tubemq-client-cpp/src/logger.cc
```diff
@@ -0,0 +1,58 @@
+
+
```
gosonzhang, line 2: need license

tubemq-client-twins/tubemq-client-cpp/example/log/main.cc
```diff
@@ -0,0 +1,28 @@
+
```
gosonzhang, line 1: need license

PR #190, opened 2020-07-06
[TUBEMQ-268] C++ SDK log module

charlely:
[TUBEMQ-268] C++ SDK log module


tubemq-client-twins/tubemq-client-cpp/example/log/main.cc
```diff
@@ -0,0 +1,28 @@
+
```
gosonzhang, line 19: need license

PR #234, opened 2020-08-10
[TUBEMQ-311] Feedback more production information

gosonzhang:


gosonzhang:
The main changes are as follows:
1. Add 3 parameters in SendMessageResponseB2P: messageId, appendTime, appendOffset, respectively save the messageId of the message storage, the time to save the message, and the index offset of the save message;
2. Adjust the generation method of messageId, which is globally unique by the broker, and adjusted to be unique within the topic: messageId only needs to be unique within the specified topic, reducing the conflict processing of data generation;
3. Modify MessageSentResult and add 2 parameters appendTime and appendOffset for the caller to use.

PR #240, opened 2020-08-19
[TUBEMQ-327] Fix the concurrency problem in the example

zjmeow:
The example was to subscribe to multiple topics in a single thread.  Now modify it to subscribe to topics on multiple threads.


gosonzhang (approved):
Thanks, you are right

PR #248, opened 2020-08-31
[TUBEMQ-316] add exception process for BindException, print log and then exit

aloyszhang:
add exception process for BindException, print log and then exit


gosonzhang (approved):
+1

lamberliu (approved):

PR #251, opened 2020-09-08
[TUBEMQ-342] Abstract backend threads for routine management

yuanboliu:


gosonzhang (approved):
+1

tubemq-manager/src/main/java/org/apache/tubemq/manager/backend/AbstractDaemon.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Stop thread pool and running threads if they're in the running state.
+     *
+     * @param timeout - max wait time
+     * @param timeUnit - time unit
+     */
+    public void waitForTerminate(long timeout, TimeUnit timeUnit) throws Exception {
+        // 停止正在执行的线程
```
aloyszhang, line 90: it's better to use english

PR #277, opened 2020-09-25
 [TUBEMQ-363]Adjust the version information of all pom.xml

lamberliu:
[TUBEMQ-363]Adjust the version information of all pom.xml


gosonzhang (approved):
+1

PR #281, opened 2020-10-03
[TUBEMQ-368]When tubemq::TubeMQConsumer shutdown() is called, there i…

charlely:
When tubemq::TubeMQConsumer shutdown() is called, there is a small probability that it will be cored directly


gosonzhang (approved):
+1

PR #282, opened 2020-10-09
[TUBEMQ-370] Calling GetCurConsumedInfo API always returns failure

gosonzhang:
[TUBEMQ-370] Calling GetCurConsumedInfo API always returns failure


charlely:
ok

PR #290, opened 2020-10-22
[TUBEMQ-380] fix static const variable definition

keyihao:
Add static const variable definition to fix link error reported at https://issues.apache.org/jira/browse/TUBEMQ-380


gosonzhang (approved):
+1

PR #304, opened 2020-11-04
[TUBEMQ-405] add cpp configurations for python sdk install

dockerzhang:
add cpp configurations for python sdk install


gosonzhang (approved):
+1

PR #321, opened 2020-11-20
[TUBEMQ-420] python client support parse message from binary data for…

dockerzhang:
[TUBEMQ-420] python client support parse message from binary data for python SDK based on pybind11


gosonzhang:
thanks

gosonzhang (approved):
+1

gosonzhang:
After I understand your implementation and reproduce it, I can only say that cross-language calls are really a big challenge.

Thanks! @dockerzhang

PR #323, opened 2020-11-22
[TUBEMQ-419]SetMaxPartCheckPeriodMs() negative number, getMessage() still lock the thread

gosonzhang:


gosonzhang:
I have annotated the two functions TubeMQConsumer.GetMessage() and ConsumerConfig.SetMaxPartCheckPeriodMs() on the header files of tubemq_client.h and const_config.h respectively, I think it will be clearly present the problem and how to use for the time being; 

The C/C++ SDK still lacks the the Producer role realization, after the subsequent processing on the Producer role is finished, I would like to produce a complete document to help you, but my English is not very good, so let’s first begin from Chinese.

lamberliu (approved):

PR #343, opened 2020-12-08
[TUBEMQ-444]Add consume and produce Cli commands

gosonzhang:
1. Restore the overwritten example codes;
2. Add Cli script commands for production and consumption
3. Adjust the naming of the Cli scripts;


lamberliu (approved):
LGTM

PR #345, opened 2020-12-09
[TUBEMQ-445]Adjust the status check default sleep interval of pullConsumeReadyChkSliceMs

gosonzhang:
Since the available status of the partition is also controlled by the server, if the detection sleep interval is too long, the partition consumption delay response will not be sensitive enough, so adjust the value to a smaller value.


lamberliu (approved):
+1

PR #346, opened 2020-12-09
[TUBEMQ-446]Small bugs fix that do not affect the main logics

gosonzhang:
Found some small bugs
1. AddressUtils.getIPV4LocalAddress() should first check whether localIPAddress has been set, and use it directly under the premise that it has been set to reduce unnecessary get operations;
2. The fetchCount parameter in MessageConsumerExample may be negative, so the executorService object should not use fetchCount as the initial value.
3. Remove the useless parameter WITHOUTDELAY


lamberliu (approved):
LGTM

lamberliu (approved):

PR #347, opened 2020-12-11
[TUBEMQ-447] Add Broker-Admin Cli

gosonzhang:
1. Add Broker-Admin's Cli command, directly enter the command(deficiency: I want to implement the command and then bring the sub-command mode, and found that commons-cli cannot support my own ideas, temporarily adopt this plan, and adjust it when there is a better plan );
2. Modify some implementation bugs of web-api


lamberliu (approved):
LGTM

PR #349, opened 2020-12-15
[TUBEMQ-451]Replace ConsumeTupleInfo with Tuple2

gosonzhang:
The current implementation of ConsumeTupleInfo can be completely replaced by Tuple2 to reduces duplicate class definitions.


yuanboliu:
+1
I rememeber apache common jar also has the implements of Tuple, not sure whether we can replace org.apache.tubemq.corebase.utils.Tuple2 with apache common Tuple

gosonzhang:
@yuanboliu 

Yes, I know, the apache common jar includes Pair and Triple classes.

In this scenario, I just do a simple return result encapsulation, without much content, and don’t want to introduce a large class for a return value encapsulation class to make the system very bloated; at the same time, I think its scheme is too rigid, Groups can be 2-tuples, 3-tuples, 4-tuples, etc, if only Pair and Triple are allowed, this is not good for subsequent program definitions.

PR #351, opened 2020-12-15
[TUBEMQ-450]TubeClientException: Generate producer id failed

gosonzhang:
1. Check the IP type when adding the local IP input setting;
2. Cancel the requirement of compulsory setting of IP address, if it does not exist, take the first IPV4 address of the machine;
3. The client configuration adds the localAddress information printing


dockerzhang (approved):
+1

PR #353, opened 2020-12-17
[TUBEMQ-458] authAuthorizedToken is always null

yuanboliu:


gosonzhang (changes requested):
This is a reserved parameter for subsequent authentication and authorization processing，and it's useful

yuanboliu:
ok, close this pr

PR #368, opened 2020-12-25
[TUBEMQ-475] add the offset clone api of the consume group

gosonzhang:
1. Adjust OffsetStorage class, add the consum group information reading APIs of zk storage;
2. Adjust OffsetService class, add query and modify APIs of group Offset;
3. Adjust BrokerAdminServlet class, add group query, group offset query, and group offset clone APIs


lamberliu (approved):
+1

PR #379, opened 2021-01-05
[TUBEMQ-490] add group \ black group \ enable auth-control \ query co…

EMsnap:
…nsumer group \ query black group

### DIFF
--- a/tubemq-manager/src/main/java/org/apache/tubemq/manager/controller/topic/TopicWebController.java
+++ b/tubemq-manager/src/main/java/org/apache/tubemq/manager/controller/topic/TopicWebController.java
@@ -18,7 +18,10 @@
 
 package org.apache.tubemq.manager.controller.topic;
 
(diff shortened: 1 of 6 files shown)


tubemq-manager/src/main/java/org/apache/tubemq/manager/controller/topic/TopicWebController.java
```diff
@@ -155,11 +162,129 @@ public TubeMQResult cloneTopic(@RequestBody CloneTopicReq req) throws Exception
      * @return
      * @throws Exception
      */
+    @PostMapping("/query/topic-config")
```
gosonzhang, line 165: Here has a  conflict with base code

yuanboliu:
+1

tubemq-manager/src/main/java/org/apache/tubemq/manager/controller/topic/TopicWebController.java
EMsnap: done

PR #380, opened 2021-01-06
[TUBEMQ-486]Add the delete API of consumer group offset

gosonzhang:
1. Add delete api of consume group offset;
2. Adjust the read implementation of ZK storage content
3. Optimized code implementation


lamberliu (approved):
LGTM

PR #389, opened 2021-01-10
[TUBEMQ-500] Add setting operate API

gosonzhang:
1. add admin_query_cluster_default_setting
2. admin_set_cluster_default_setting

Currently, these only supports the setting of maxMsgSize, and later we can support other parameter settings based on this structure.


yuanboliu:
+1

PR #395, opened 2021-01-13
[TUBEMQ-506] cmakelist error, rm -Werror

myd620:
fix compile error


charlely:
You can use -Wno-error=class-memaccess , to close class-memaccess warning, but cann't close all warn to error.

charlely:
agree

gosonzhang (approved):
+1

PR #396, opened 2021-01-13
[TUBEMQ-511]Replace the conditional operator (?:) with mid()

gosonzhang:
At present, many places in the code implementation use conditional operators(? : ) for value checking, but it's not very readable, and the MixedUtils.mid() function is used instead


yuanboliu:
+1 pending on test result.

PR #405, opened 2021-01-20
[TUBEMQ-530] create a cluster in manager

EMsnap:
[TUBEMQ-530] create a cluster in manager


tubemq-manager/pom.xml
```diff
@@ -54,6 +54,12 @@
             <version>4.3</version>
         </dependency>
 
+        <dependency>
```
yuanboliu, line 57: seems useless in this patch

tubemq-manager/src/main/java/org/apache/tubemq/manager/entry/NodeEntry.java
```diff
@@ -32,8 +32,8 @@
 @Data
 public class NodeEntry {
     @Id
-    @GeneratedValue(strategy= GenerationType.AUTO)
-    private long brokerId;
+    @GeneratedValue(strategy= GenerationType.IDENTITY)
```
yuanboliu, line 35: what if clusters have conflicted broker ids. 
for example A cluster has broker which id is num_a, meanwhile B cluster also has broker with the same id.
It seems better that we need create a colume called brokerId without any annotation.

tubemq-manager/src/main/java/org/apache/tubemq/manager/entry/ClusterEntry.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@Entity
+@Table(name = "cluster")
+@Data
+public class ClusterEntry {
+    @Id
+    @GeneratedValue(strategy= GenerationType.IDENTITY)
+    private int clusterId;
+
+    private String clusterName;
```
yuanboliu, line 40: cluster name should be marked as uniq key.

tubemq-manager/pom.xml
EMsnap: 21.0

yuanboliu:
+1

PR #410, opened 2021-01-28
[TUBEMQ-546]Restore the original license header of the referenced external source files

gosonzhang:
1. Restore the original license header of the referenced external source files;
2. Modify rat configuration in pom.xml to exclude the LICENSE normative check of the cited external source files;
3. Confirm that the referenced source file already has a LICENSE statement(already exists, no need to modify).


dockerzhang (approved):
+1

PR #422, opened 2021-02-03
[TUBEMQ-556]Index value is bigger than the actual number of records

gosonzhang:
1. Accumulation rules for alignment record numbers;
2. Increase the space between different content in log printing


yuanboliu:
+1

PR #423, opened 2021-02-08
[TUBEMQ-555] short session data can only be written to a specific partition (addendum)

gosonzhang:
Co-authored-by: 曹显乐 <xianle.cao@vivo.com>


yuanboliu:
+1

PR #424, opened 2021-02-08
[TUBEMQ-558] Adjust the LICENSE of the file header

gosonzhang:
The following files do not carry the standard Apache V2 LICENSE header:

tubemq-docker/tubemq-all/Dockerfile
tubemq-manager/bin/start-manager.sh
tubemq-manager/bin/stop-manager.sh
tubemq-manager/conf/logback.xml
tubemq-manager/src/main/assembly/assembly.xml


yuanboliu:
+1

PR #426, opened 2021-02-11
[TUBEMQ-560] Remove unprepared modules

gosonzhang:
1. There is some problems with the pom.xml of tubemq-docker, which affects the project compilation, and will be released after adjustment;

2. All files in the tubemq-web directory are missing LICENSE header and do not meet the publishing conditions. Delete them now and publish them after they are ready;


yuanboliu:
+1

PR #427, opened 2021-02-11
[TUBEMQ-559] Update the LICENSE file according to the 0.8.0-RC2 review [addendum]

gosonzhang:
1. Add the LICENSE informations of binary dependency packages;
2. Remove tubem-manager module for LICENSE reasons.


yuanboliu:
+1

PR #438, opened 2021-03-03
[TUBEMQ-566] manage brokers in region

EMsnap:
[TUBEMQ-566] manage brokers in region


yuanboliu (commented):
It would be great if we could add some unit tests for this change.

EMsnap:
> It would be great if we could add some unit tests for this change.

would be done soon, thanks

yuanboliu (approved):

yuanboliu:
+1

PR #443, opened 2021-03-12
[TUBEMQ-576] Build metadata entity classes

gosonzhang:
1. Increase the representation object of metadata in memory;
2. Add operation interfaces that need to be supported by the actual storage layer of metadata
3. Other minor adjustments and optimizations


lamberliu (approved):
+1

PR #445, opened 2021-03-17
[TUBEMQ-579] Add structure mapping of BDB and metadata entity classes

gosonzhang:
1. Add APIs to convert BDB storage objects into corresponding metadata objects, convert metadata objects into BDB storage objects;
2. Adjust the current implementation of BDB and add the extension fields operation API;
3. Adjust the Token definition of the extension fields and summarize it into one file;


dockerzhang (approved):
+1

PR #450, opened 2021-03-30
[TUBEMQ-584] Adjust WebMasterInfoHandler class implementation

gosonzhang:
1. Add clone() and toWebJsonStr() functions to all dao classes
2. Adjust the abbreviations of existing fields in WebFieldDef and add new fields
3. Refactor the implementation of the WebMasterInfoHandler class according to the new metadata class


dockerzhang (approved):
+1

PR #458, opened 2021-04-21
[INLONG-599]Adjust WebParameterUtils.java's static functions

gosonzhang:
1. Adjust the corresponding functions, changing the field container parameter type to generic type
2. Adjust other implementations


yuanboliu:
+1

PR #461, opened 2021-04-29
[INLONG-601]Adjust WebBrokerDefConfHandler class implementation

gosonzhang:
1. Replace the existing web api implementation, adopt a new metadata management for data interaction;
2. Adjust the web api call plan, and pass in the common parameters from the caller.


dockerzhang (approved):
+1

PR #475, opened 2021-05-26
[INLONG-618] Add unit tests for metastore.dao.entity.*

gosonzhang:
add unit test cases for metastore.dao.entity.*


dockerzhang (approved):
+1

PR #496, opened 2021-07-05
[INLONG-637]Go SDK log

TszKitLo40:
PTAL @charlely @gosonzhang


gosonzhang (approved):
+1

tubemq-client-twins/tubemq-client-go/log/log.go
```diff
@@ -0,0 +1,127 @@
...
+
+// package log defines the logger for the sdk.
+package log
+
+import (
+	"sync"
+)
+
+func init() {
+	defaultLogger = newDefaultLogger()
```
charlely, line 26: Since the defaultlog is built in init, can it be built globally to reduce the dependency on once?

charlely (approved):
LGTM

PR #554, opened 2021-07-16
[INLONG-750] Improve master.ini and broker.ini

LeBW:
1. change default port to 8715 in master.ini
2. add configuration instruction for defEthName in broker.ini

### DIFF
--- a/inlong-tubemq/conf/master.ini
+++ b/inlong-tubemq/conf/master.ini
@@ -19,7 +19,7 @@
 ; host address of master, required; must be configured at network card, enabled
 hostName=127.0.0.1
 ; port that master listens to, optional; default is 8715
-port=8000
+port=8715
 ; port that master web console listens to
 webPort=8080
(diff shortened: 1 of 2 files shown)


inlong-tubemq/conf/master.ini
```diff
@@ -19,7 +19,7 @@
 ; host address of master, required; must be configured at network card, enabled
 hostName=127.0.0.1
 ; port that master listens to, optional; default is 8715
-port=8000
+port=8715
```
gosonzhang, line 22: Does the modification to 8715 have any special meaning?

LeBW: As shown in the comment in Line 21, the default port is 8715. Besides, the default port in `MasterConfig` is 8715, too.
So maybe it's better to use 8715 here, so it won't confuse users.

gosonzhang: Ok

gosonzhang (approved):
+1

PR #1574, opened 2021-09-19
[INLONG-1573][TubeMQ] Add TDMsg decode logic to TubeMQ's C++ SDK

### LINKED ISSUE #1573: Add TDMsg decode logic to TubeMQ's C++ SDK
When data is consumed, the consumer of InLong needs to decode the message in the TDMsg format, so as to obtain the correct reported data

This modification solves the problem.

gosonzhang:
Fixes #1573


dockerzhang (approved):
+1

PR #1576, opened 2021-09-19
[INLONG-1568][TubeMQ] C++ SDK cant not return the whether the partition has been registered correctly

### LINKED ISSUE #1568: C++ SDK cant not return the whether the partition has been registered correctly
#### Expected behavior

Tell us what should happen
C++ SDK cant not return the whether the partition has been registered correctly
#### Actual behavior

Tell us what happens instead

https://github.com/apache/incubator-inlong/blob/4b2223269da92aacd0cec8d40da4d624dc42ca3e/inlong-tubemq/tubemq-client-twins/tubemq-client-cpp/src/rmt_data_cache.cc#L458

In the function `IsPartitionFirstReg`, if `part_reg_booked_` can't find the partition, it will add the partition to the map and return true. And if the function is called second time, it will also return true because the status has been set to true.
#### Steps to reproduce

How can we reproduce the issue

#### System configuration
**InLong version**: 0.11.0

gosonzhang:
Fixes #1568


dockerzhang (approved):
+1

Issue #1579, opened 2021-09-22
 lots of files are not standard License Header


dockerzhang:
#### Expected behavior

from #1572 , we found lots of files are not standard License Header.

#### Actual behavior

Tell us what happens instead

#### Steps to reproduce

How can we reproduce the issue

#### System configuration
**InLong version**: x.y

PR #1582, opened 2021-09-22
[INLONG-1579] Fix the license header error in cmd and conf files

shink:
issue: #1579

### Motivation

Fix the license header error in cmd and conf files.

BTW, the license header in `cmd` type files should start with '@rem' instead of `rem` and in `conf` type files should start with `#` instead of `;`.


shink:
According to https://github.com/apache/incubator-inlong/pull/1572, only `.ini` files cann't pass the check now.

dockerzhang:
@gosonzhang PTAL

dockerzhang (approved):

PR #1698, opened 2021-10-20
[INLONG-1670][Manager] Revert charset change of table

ruanwenjun:
This pr recovers the error deletion change of charset in #1671


ruanwenjun:
@healchow Please take a look.

PR #1773, opened 2021-11-09
[INLONG-1772][Improve]Adjust the ProcessResult class implementation

### LINKED ISSUE #1772: [Improve]Adjust the ProcessResult class implementation
### Description

The ProcessResult class is not only used in the Server module but also used on the client-side, putting it in the core module can play a better role; at the same time, the member variables defined by this class are not read through functions at some points of use, and the code is written not very good.

So this issue makes corresponding modifications here to meet the above requirements.

### InLong Component

InLong TubeMQ

### Code of Conduct

gosonzhang:
Fixes #1772


aloyszhang (approved):
+1

PR #1918, opened 2021-12-07
[INLONG-1911] add pulling metric of prometheus, pulling metric value will not be resetted.

### LINKED ISSUE #1911: [Bug] Some questions about the metric implementation in the common module
### What happened

I analyzed the metric implementation in the common module[1], and there are a few issues that need to be corrected, or they can’t be connected with the external standard JMX call.
a. To be implemented based on MXBean instead of MBean;
    MBean can only provide the call of the basic class, and the composite class needs MXBean to define
b. The name format of the constructed Beanobject should be unified:
    To be constructed based on ObjectName("XXXX:type=YYYYY") format, where XXXX and YYYYYare arbitrary, but ":type=" is a fixed format
c. Cannot only have get-and-reset() API:
    If we only have the get-and-reset() API, when we look at Counter type indicators, we still need the caller to keep the previous data and accumulate it by itself, which is unreasonable.

1. https://github.com/apache/incubator-inlong/tree/master/inlong-common/src/main/java/org/apache/inlong/commons/config/metrics

Use standard JMX tools, for example, jconsole, for interface access

### How to reproduce

Use standard JMX tools, for example, jconsole, for interface access

### Environment

_No response_

### InLong version

master

### InLong Component

Other for not specified component

### Code of Conduct

luchunliang:
[INLONG-1911] add pulling metric of Prometheus, pulling metric value will not be reset.

### Title Name: [INLONG-1911][DataProxy]  add pulling metric of Prometheus, pulling metric value will not be reset.

Fixes #1911

### Motivation

*Explain here the context, and why you're making that change. What is the problem you're trying to solve.*

### Modifications

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*

### Documentation

  - Does this pull request introduces a new feature? (yes / no)

### DIFF
--- a/inlong-dataproxy/dataproxy-source/pom.xml
+++ b/inlong-dataproxy/dataproxy-source/pom.xml
@@ -26,8 +26,8 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<compiler.source>1.8</compiler.source>
 		<compiler.target>1.8</compiler.target>
-		<junit.version>4.13</junit.version>
 		<guava.version>19.0</guava.version>
+		<powermock.version>2.0.9</powermock.version>
 		<skipTests>false</skipTests>
 	</properties>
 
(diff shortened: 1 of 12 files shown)


aloyszhang:
@baomingyu @EMsnap PTAL

gosonzhang (approved):
+1

dockerzhang (approved):

inlong-dataproxy/dataproxy-source/pom.xml
```diff
@@ -46,19 +46,13 @@
 		<dependency>
 			<groupId>org.powermock</groupId>
 			<artifactId>powermock-module-junit4</artifactId>
-			<version>2.0.2</version>
+			<version>${powermock.version}</version>
```
inter12, line 49: The third-party dependent version needs to be external to the parent pom  <dependencymanagement> file

inter12: dependencymanagement

luchunliang: ok, move dependency to pom.xml of inlong-dataproxy

PR #2208, opened 2022-01-20
[INLONG-2191][AuditSource] fix requestId is not set in response message

baomingyu:
fix requestId is not set in response message

Fixes # 2191


aloyszhang (approved):
LGTM

dockerzhang (approved):

PR #2415, opened 2022-02-09
[INLONG-2414][Manager] Exclude test jar file during the apache-rat-plugin check

### LINKED ISSUE #2414: [Manager] Exclude test jar file during the apache-rat-plugin check
### Description

Exclude test jar file during the apache-rat-plugin check.

### InLong Component

InLong Manager

### Code of Conduct

healchow:
### Title Name: [INLONG-2414][Manager] Exclude test jar file during the apache-rat-plugin check

Fixes #2414

### Motivation

Exclude test jar file during the apache-rat-plugin check.

### Modifications

Add test into excluding config.

### Verifying this change

This change is a trivial rework/code cleanup without any test coverage.

### Documentation

  - Does this pull request introduces a new feature? no


binaryWorld (approved):

dockerzhang (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/plugin/PluginService.java
```diff
@@ -50,7 +50,7 @@
     private List<PluginBinder> pluginBinders;
 
     @Getter
-    private List<Plugin> plugins = new ArrayList<>();
+    private final List<Plugin> plugins = new ArrayList<>();
```
aloyszhang, line 53: for uniqueness, maybe using a Set<Plugin> here is better

healchow: Each plugin instance was created by `new`, and not overwrite the `toString` and `equals` methods, so `Set` and `List` are the same in this scenario.

aloyszhang (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/plugin/PluginService.java
aloyszhang: @kipshi Could you confirm that all the plugins here should be a singleton?
If so, current logic may break this semantic, e.g. if we have two `PluginDefinitions` with the same `pluginClass` and different `name`, the `List<Plugin> plugins` will have two instances for the same `pluginClass`.

PR #2580, opened 2022-02-18
[INLONG-2569][TubeMQ] Discarded msgTime value when msgType is empty

### LINKED ISSUE #2569: [TubeMQ] Discarded msgTime value when msgType is empty
### What happened

See https://github.com/apache/incubator-inlong/blob/master/inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corebase/utils/MixedUtils.java#L125
image

In the MixedUtils.buildMessage() function, there is a bug hidden in the above part: when the filter item does not exist, the message reporting time information will not be filled in the message, although these two parameters are filled in through the same API, analyze the implementation of putSystemHeader, It is a setting that supports the case where two parameters are null at the same time. Therefore, it is necessary to adjust here to avoid the situation of missing information.

send message without filter item

### How to reproduce

send message without filter item

### Environment

windows

### InLong version

master

### InLong Component

InLong TubeMQ

### Code of Conduct

gosonzhang:
Fixes #2569


dockerzhang (approved):

aloyszhang (approved):
+1

Issue #2632, opened 2022-02-21
[Bug][K8s][Pulsar] The pulsar should be an external component on Kubernetes


shink:
### What happened

The pulsar is not an component of InLong, it should be an external component on Kubernetes.

### What you expected to happen

The pulsar component on Kubernetes can be configured by the user.

### How to reproduce

Now the pulsar is an component on Kubernetes.

### InLong version

master

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2633, opened 2022-02-21
[INLONG-2632][K8S] Add support for pulsar configuration

shink:
Fix: #2632

### Motivation

The pulsar is not an component of InLong, it should be an external component on Kubernetes and can be configured by the user.

### Modifications

- `docker/kubernetes`

### Verifying this change

- [x] Make sure that the change passes the CI checks.


dockerzhang:
@shink The pulsar is not a component of InLong, therefore, we should remove all pod configurations for it:
image

shink:
@dockerzhang Done.

healchow (approved):

aloyszhang (approved):

PR #2639, opened 2022-02-21
[INLONG-2638][Manager] apache_inlong_manager.sql file execution exception

Greedyu:
https://inlong.apache.org/docs/next/modules/manager/quick_start . initialize database there is sql/apache_inlong_manager.sql. Throw an exception: ERROR 1364 (HY000) at line 316: Field 'mq_set_name' doesn't have a default value

### Title Name: 
[INLONG-2638][Manager] apache_inlong_manager.sql file execution exception
Fix: https://github.com/apache/incubator-inlong/issues/2638

### Motivation

Failed to initialize database

### Modifications

Because the subsequently inserted instance does not assign a value to the mq_set_name field, the default value should be nullable


Greedyu:
already fixed

healchow:
See this PR: https://github.com/apache/incubator-inlong/pull/2656

Issue #2646, opened 2022-02-22
[Bug][MQ] start_inlong_all function start problem


Greedyu:
### What happened

When starting inlong, the message middleware selects pulsar, but the script does not detect it. 
![wecom-temp-7e9f791b799e5f97c68f4f40c9467772](https://user-images.githubusercontent.com/20356765/155068728-a333d2d2-aaa5-462e-a935-af33df354339.png)

because the code makes judgments based on the PulsarBrokerStarter startup class. And pulsar has two startup classes, PulsarBrokerStarter and PulsarStandaloneStarter
<img width="852" alt="image" src="https://user-images.githubusercontent.com/20356765/155068767-2d13cebc-b658-414c-adc7-ab020cd352bb.png">

### What you expected to happen

The detection of the pulsar process should be coupled with another startup class PulsarStandaloneStarter

### How to reproduce

The message middleware uses pulsar in standalone mode, and then starts inlong

### Environment

centos

### InLong version

master

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #2647, opened 2022-02-22
[INLONG-2646][MQ]The startup script cannot detect that pulsar starts the process in standalone mode

Greedyu:
### Title Name: [INLONG-2646][MQ] The startup script cannot detect that pulsar starts the process in standalone mode

Fixes [#2646](https://github.com/apache/incubator-inlong/issues/2646)

### Motivation

When starting inlong, the message middleware selects pulsar, but the script does not detect it.

### Modifications

the detection of the pulsar process adds another startup class PulsarStandaloneStarter


dockerzhang (approved):

aloyszhang (approved):
+1

PR #2664, opened 2022-02-22
[INLONG-2662][Manager] Fix bug to avoid reinit of service task

### LINKED ISSUE #2662: [Bug] Fix duplicate listener in manager service
### What happened

When init workflow definition, in some cases,  duplicate listener will be added in service task;

When init workflow definition, in some cases,  no duplicate listener will be added in service task;

### How to reproduce

Create workflow twice , than will find exception be throwed;

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

kipshi:
### Title Name: [INLONG-2662][Manager] Fix bug to avoid reinit of service task

Fixes #2662 

### Verifying this change

This change is a trivial rework/code cleanup without any test coverage.

### Documentation

  - Does this pull request introduces a new feature? no

### DIFF
--- a/inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/definition/ServiceTask.java
+++ b/inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/definition/ServiceTask.java
@@ -19,6 +19,7 @@ package org.apache.inlong.manager.workflow.definition;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
+import java.util.concurrent.atomic.AtomicBoolean;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.inlong.manager.common.exceptions.WorkflowException;
 import org.apache.inlong.manager.common.util.Preconditions;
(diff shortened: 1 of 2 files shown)


inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/definition/ServiceTask.java
```diff
@@ -42,6 +44,8 @@
 
     private ServiceTaskType serviceTaskType;
 
+    private boolean isInit;
```
gosonzhang, line 47: For isInit

If there is multi-threaded access, it is recommended to add volatile define. If there is a multi-threaded concurrent modification, at least a lock is required to avoid reentrancy

kipshi: Excellent comment !!

kipshi: I prefer to use CAS to avoid multi-thread modification like this;

private AtomicBoolean isInit = new AtomicBoolean(false);

public void initListeners(WorkflowContext workflowContext) {
        if (isInit.compareAndSet(false, true)) {
            if (listenerProvider == null || serviceTaskType == null) {
                return;
            }
            Iterable<TaskEventListener> listeners = listenerProvider.get(workflowContext, serviceTaskType);
            addListeners(Lists.newArrayList(listeners));
        } else {
            log.debug("ServiceTask:{} is already init", getName());
        }
    }

healchow (approved):

dockerzhang (approved):

PR #2724, opened 2022-02-25
[INLONG-2723][Manager] Fix batch insert entities error in AbstractService

### LINKED ISSUE #2723: [Bug][Manager] Manager module occurred exception when startup
### What happened

manager will thread exception while starting
wecom-temp-bbb728ff4cd82e2384f7a9dfd079acec

manager will thread exception while starting

### How to reproduce

manager will thread exception while starting

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

baomingyu:
### Title Name: [INLONG-2723][Manager] Fix batch insert entities error in AbstractService

Fixes #2723

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AbstractService.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AbstractService.java
@@ -24,6 +24,7 @@ import java.util.concurrent.LinkedBlockingDeque;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import org.apache.commons.collections.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AbstractService.java
```diff
@@ -102,7 +102,9 @@ public  void run() {
                         }
                         count++;
                     }
-                    batchInsertEntities(entryList);
+                    if (entryList.size() > 0) {
```
ciscozhou, line 105: Its recommanded use `CollectionUtils.isNotEmpty` to check whether a collection is empty or not.

baomingyu: CollectionUtils.isNotEmpty  need check more than once.
like this :
coll == null || coll.isEmpty();
 and here entryList is never be null. so check entryList.size()  will be beset.

healchow (approved):

dockerzhang (approved):

PR #2762, opened 2022-02-26
[INLONG-2759][INLONG-2760][Manager] Fix InlongGroup object deletion and modification exception

### LINKED ISSUE #2759: [Bug][Manager] InlongGroupController update interface status problem
### What happened

First create a group to be submitted.
image
Enter the page by clicking Details and click Next. An exception will occur：There was an error in the service...Please try again later! If there are still problems, please contact the administrator
image

There should be no exception

### How to reproduce

has been described earlier

### Environment

centos

### InLong version

master

### InLong Component

InLong Manager, InLong Dashboard

### Code of Conduct

### LINKED ISSUE #2760: [Bug][Manager] Delete data grouping exception
### What happened

image

There should be no exception

### How to reproduce

Just delete an InlongGroup

### Environment

centos

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

Greedyu:
### Title Name: [INLONG-2759][INLONG-2760][Manager] Fix InlongGroup object deletion and modification exception

Fixes #2759  Manager module
First of all, the reason is that the GroupState.forCode() function has passed a null value. The reason is that the request object InlongGroupRequest does not add the annotation @ApiModelProperty() to the status field, which has been fixed here.

But there is still a problem with retesting with annotations, because the front-end page does not pass in the status parameter.
image

Then continue to use the postman tool to form the status parameter for testing. A new problem has arisen: Current state=wait_submit is not allowed to transfer to state=wait_submit
image

This is because there is no conversion from GROUP_WAIT_SUBMIT to GROUP_WAIT_SUBMIT in the status transition. The above scenario should be missed here, so I fixed it and tested it again through the post. Finally verified successfully

image

Finally, the front-end Dashborad module needs to be repaired and the status parameter added.

Fixes #2760 
Because the deletion actually calls the groupProcessOperation.deleteProcess() function, the InlongGroupRequest queried according to the id does not change the state. In the inner layer, the groupService.update(groupInfo, username) method is actually executed first. equivalent to the original state of transformation
image

After repairing and testing again, the interface can return successfully
image


healchow:
It is not recommended to associate 2 or more issues in one PR.

PR #2769, opened 2022-02-27
[INLONG-2759][Manager] Fix InlongGroup object  modification exception

### LINKED ISSUE #2759: [Bug][Manager] InlongGroupController update interface status problem

Greedyu:
### Title Name: [INLONG-2759][Manager] Fix InlongGroup object  modification exception

Fixes #2759 

First of all, the reason is that the GroupState.forCode() function has passed a null value. The reason is that the request object InlongGroupRequest does not add the annotation @ApiModelProperty() to the status field. 

Originally, an annotation was added to the status field, but the reviewer believes that the status should not be set in the request, and there may be abnormal status flow caused by front-end parameters. We should handle state in the backend system.  
image

Then continue to use the postman tool to form the status parameter for testing. A new problem has arisen: Current state=wait_submit is not allowed to transfer to state=wait_submit
image

This is because there is no conversion from GROUP_WAIT_SUBMIT to GROUP_WAIT_SUBMIT in the status transition. The above scenario should be missed here, so I fixed it and tested it again through the post. Finally verified successfully

image

The state needs to be controlled in the interface, and the background business logic is required later here.


inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/enums/GroupState.java
```diff
@@ -46,7 +46,7 @@
         GROUP_FINITE_STATE_AUTOMATON.put(GROUP_DRAFT,
                 Sets.newHashSet(GROUP_WAIT_SUBMIT, GROUP_DELETE));
         GROUP_FINITE_STATE_AUTOMATON.put(GROUP_WAIT_SUBMIT,
-                Sets.newHashSet(GROUP_WAIT_APPROVAL, GROUP_DELETE));
+                Sets.newHashSet(GROUP_WAIT_SUBMIT, GROUP_WAIT_APPROVAL, GROUP_DELETE));
```
kipshi, line 49: this has been fixed in master

Greedyu:
has been fixed

PR #2770, opened 2022-02-27
[INLONG-2760][Manager] Fix InlongGroup object deletion

### LINKED ISSUE #2760: [Bug][Manager] Delete data grouping exception

Greedyu:
### Title Name: [INLONG-2760][Manager] Fix InlongGroup object deletion

Fixes #2760 
Because the deletion actually calls the groupProcessOperation.deleteProcess() function, the InlongGroupRequest queried according to the id does not change the state. In the inner layer, the groupService.update(groupInfo, username) method is actually executed first. equivalent to the original state of transformation

image

After repairing and testing again, the interface can return successfully
image


healchow:
Was fixed in `UpdateGroupCompleteListener` class.

healchow:
Was fixed in `UpdateGroupCompleteListener` class.

PR #3356, opened 2022-03-24
[INLONG-3355][Manager] Check the source states before return group state in manager client

### LINKED ISSUE #3355: [Improve] Check sourceState When return GroupState in Manager Client
### Description

_No response_

### InLong Component

InLong Manager

### Code of Conduct

kipshi:
### Title Name: [INLONG-3355] Check the source states before return group state in manager client

Fixes #3355

### Verifying this change

### DIFF
--- a/inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
+++ b/inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
@@ -122,7 +122,7 @@ public class InlongGroupContext implements Serializable {
         switch (this.state) {
             case STARTED:
                 for (StreamSource source : sourcesInGroup) {
-                    if (!(source.getState() == State.NORMAL)) {
+                    if (source.getState() != State.NORMAL) {
                         log.warn("StreamSource:{} is not started", source);
                         this.state = InlongGroupState.INITIALIZING;
                         break;
@@ -131,7 +131,7 @@ public class InlongGroupContext implements Serializable {
                 return;
             case STOPPED:
                 for (StreamSource source : sourcesInGroup) {
-                    if (!(source.getState() == State.FROZEN)) {
+                    if (source.getState() != State.FROZEN) {
                         log.warn("StreamSource:{} is not stopped", source);
                         this.state = InlongGroupState.OPERATING;
                         break;
--- a/inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
+++ b/inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
@@ -20,6 +20,7 @@ package org.apache.inlong.manager.client.api;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.inlong.manager.client.api.StreamSource.State;
@@ -35,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 
 @Data
+@Slf4j
 public class InlongGroupContext implements Serializable {
 
     private String groupId;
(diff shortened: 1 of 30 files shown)


gong (approved):

crancchen (approved):

dockerzhang (approved):

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/pojo/stream/InlongStreamResponse.java
```diff
@@ -85,29 +84,6 @@
     @ApiModelProperty(value = "Names of responsible persons, separated by commas")
     private String inCharges;
 
-    @ApiModelProperty(value = "Status")
```
healchow, line 88: The response needs to show those fields for the user.

inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
```diff
@@ -91,24 +93,53 @@ private void recheckState() {
...
                 this.groupErrLogs.computeIfAbsent("failedSources", Lists::newArrayList)
                         .add(GsonUtil.toJson(failedSource));
             }
+            return;
+        }
+        // check if any stream source is in indirect state
+        switch (this.state) {
+            case STARTED:
+                for (StreamSource source : sourcesInGroup) {
+                    if (!(source.getState() == State.NORMAL)) {
```
healchow, line 125: Why not use `source.getState() != State.NORMAL`?

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/source/AbstractStreamSourceOperation.java
```diff
@@ -126,11 +126,6 @@ public void updateOpt(SourceRequest request, Integer groupStatus, String operato
         // Setting updated parameters of stream source entity.
         setTargetEntity(request, entity);
         entity.setVersion(entity.getVersion() + 1);
-        if (GroupState.forCode(groupStatus).equals(GroupState.CONFIG_SUCCESSFUL)) {
```
healchow, line 129: Why remove those lines?

inlong-manager/manager-client/src/main/java/org/apache/inlong/manager/client/api/InlongGroupContext.java
kipshi: done

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/source/AbstractStreamSourceOperation.java
healchow: If the source was normal or failed, after the update operation, the status should be transfer to `new` or `to_be_add`.

kipshi: if the source was frozen, then you change to new or to_be_add, the status is fuzed

kipshi: change like this
 // If failed , return to new so can be init again
        if (SourceState.forCode(entity.getStatus()) == SourceState.SOURCE_FAILED){
            entity.setStatus(SourceState.SOURCE_NEW.getCode());
        }

healchow: Should we need concerned about the group status?
For example, if the Group was SUCCESS or FAILED, how to control the Source status?

healchow: We can handle it like this: do not modify the state when modifying the source info, and then call the restart interface after the modification.

kipshi: done

healchow (approved):
LGTM

PR #3384, opened 2022-03-27
[INLONG-3383][Manager] Fix the null pointer caused by sink field not configured with source field

### LINKED ISSUE #3383: [Bug][Manager] Fix the null pointer caused by sink field not configured with source field
### What happened

A null pointer exception will occur  if the sink field is not configured with the source field information when distributing the sort configuration.

It is expected that the sort configuration can be delivered normally, even if the sink field is not configured with source field information.

### How to reproduce

The sink field does not configure the source field information when distributing the sort configuration.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

yunqingmoswu:
### Title Name: [INLONG-3383][Manager] Fix the null pointer caused by sink field not configured with source field

Fixes #3383 

### Motivation

*Explain here the context, and why you're making that change. What is the problem you're trying to solve.*

### Modifications

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

This change is a trivial rework / code cleanup without any test coverage.

*(or)*

This change is already covered by existing tests, such as *(please describe tests)*.

*(or)*

This change added tests and can be verified as follows:

*(example:)*

### Documentation

  - If a feature is not documented yet in this PR, please create a followup issue for adding the documentation

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/thirdparty/sort/util/FieldInfoUtils.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/thirdparty/sort/util/FieldInfoUtils.java
@@ -87,10 +87,11 @@ public class FieldInfoUtils {
             FieldInfo sinkField = getFieldInfo(field.getFieldName(), field.getFieldType(),
                     field.getIsMetaField() == 1, field.getFieldFormat());
             sinkFields.add(sinkField);
-
-            FieldInfo sourceField = getFieldInfo(field.getSourceFieldName(),
-                    field.getSourceFieldType(), field.getIsMetaField() == 1, field.getFieldFormat());
-            mappingUnitList.add(new FieldMappingUnit(sourceField, sinkField));
+            if (StringUtils.isNotBlank(field.getSourceFieldName())) {
+                FieldInfo sourceField = getFieldInfo(field.getSourceFieldName(),
+                        field.getSourceFieldType(), field.getIsMetaField() == 1, field.getFieldFormat());
+                mappingUnitList.add(new FieldMappingUnit(sourceField, sinkField));
+            }
         }
 
         return mappingUnitList;


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/thirdparty/sort/util/FieldInfoUtils.java
```diff
@@ -87,10 +87,11 @@
             FieldInfo sinkField = getFieldInfo(field.getFieldName(), field.getFieldType(),
                     field.getIsMetaField() == 1, field.getFieldFormat());
             sinkFields.add(sinkField);
-
-            FieldInfo sourceField = getFieldInfo(field.getSourceFieldName(),
-                    field.getSourceFieldType(), field.getIsMetaField() == 1, field.getFieldFormat());
-            mappingUnitList.add(new FieldMappingUnit(sourceField, sinkField));
+            if (StringUtils.isNotBlank(field.getSourceFieldName())) {
```
healchow, line 90: The source and sink fields cannot be empty here. If they are empty, it means that there is a problem with the configuration, and an exception should be thrown.

healchow: Perhaps we should do non-null checks when saving and modifying source and sink fields.

yunqingmoswu: Writing a data source like hive requires all table field information, but there are some field mapping scenarios, such as 8 fields on the user source side and 10 fields on the target side. At this time, all the fields on the target side need to be passed in. But only part of the field mapping can be passed in.

healchow: OK, then the Sort module needs to handle this scenario.

healchow (approved):

dockerzhang (approved):

Issue #3744, opened 2022-04-16
[Bug][CI] Docker images are not pushed in release-* branches


shink:
### What happened

The Docker images have not be pushed in release-* branches. They should also be pushed.

### What you expected to happen

Docker images should be pushed in release-* branches.

### How to reproduce

Check out https://github.com/apache/incubator-inlong/actions/workflows/ci_build_docker.yml

### InLong version

- master
- release-*

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3748, opened 2022-04-16
[INLONG-3744][CI] Fix trigger condition when pushing Docker image

shink:
issue: #3744

### Motivation

Docker images are not pushed in `release-*` branches.

### Modifications

- Fix trigger condition when pushing Docker image
- Rename workflow: `ci_build_docker` -> `ci_docker`

BTW, unfortunately, there is any way to do regex matching on `if` conditional expressions yet on GitHub Actions. I referenced https://stackoverflow.com/questions/58862864/github-actions-ci-conditional-regex

### Verifying this change

### DIFF
--- a/.github/workflows/ci_docker.yml
+++ b/.github/workflows/ci_docker.yml
@@ -86,7 +86,7 @@ jobs:
           CI: false
 
       # If the changes are being pushed to the master branch or a branch like 'release-1.0.0', this step will output true.
-      - name: Match branch
+      - name: Match Branch
         id: match-branch
         if: |
           success()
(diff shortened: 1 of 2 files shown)


healchow (approved):

.github/workflows/ci_docker.yml
```diff
@@ -85,8 +85,22 @@ jobs:
         env:
           CI: false
 
+      # If the changes are being pushed to the master branch or a branch like 'release-1.0.0', this step will output true.
+      - name: Match branch
```
dockerzhang, line 89:
```suggestion
      - name: Match Branch
```

dockerzhang (approved):

PR #3966, opened 2022-04-27
Fix docker-compose quickstart start cmd error

xunfeng1980:
### Modifications

Fix docker-compose quickstart start cmd error

### Verifying this change

### Documentation


dockerzhang:
@xunfeng1980 thanks for your contribution
1, here we should remove the step to copy SQL files, because all of them are packed to the binary package file. you can see
https://github.com/xunfeng1980/incubator-inlong/blob/master/inlong-distribution/distribution.xml

2, the latest guide for Docker Compose, you can see https://inlong.apache.org/docs/next/deployment/docker

3, please follow the [committing guide](https://inlong.apache.org/community/how-to-commit) to modify this PR, thanks.

PR #4041, opened 2022-04-29
[INLONG-4040][Manager] Fix hive sink table data path

### LINKED ISSUE #4040: [Bug][Manager] Hive sink table path config error
### What happened

The hive table hdfs path written by sort job does not match with the default database location.
For example the hive table created by manager is /user/hive/warehouse/{db_name}.db/{table_name}.
But the sort job writes to /user/hive/warehouse/{table_name} so query in hive doesn't work.

Correct sort sink path so that data can be correctly retrieved by hive.

### How to reproduce

Configure a hive sink group and run the stream.

### Environment

MacOs

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
Fixes #4040
Tested in local env.

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sort/util/SinkInfoUtils.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sort/util/SinkInfoUtils.java
@@ -56,6 +56,7 @@ public class SinkInfoUtils {
     private static final String DATA_FORMAT = "yyyyMMddHH";
     private static final String TIME_FORMAT = "HHmmss";
     private static final String DATA_TIME_FORMAT = "yyyyMMddHHmmss";
+    private static final String HIVE_DEFAULT_DB = "default";
     private static final String HIVE_DB_SUFFIX = ".db";
 
     /**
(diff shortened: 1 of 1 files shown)


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sort/util/SinkInfoUtils.java
```diff
@@ -182,13 +183,15 @@ private static HiveSinkInfo createHiveSinkInfo(HiveSinkResponse hiveInfo, List<F
             }).collect(Collectors.toList());
         }
 
-        // dataPath = dataPath + / + tableName
-        StringBuilder dataPathBuilder = new StringBuilder();
+        // dataPath = dataPath + / + dbName + / + tableName
```
healchow, line 186: We use the data path and table name as the real path to writing files.
If changed, maybe occurred an exception in other cases.

healchow: Maybe you can change the usage of the dataPath when creating Hive tables.

woofyzhao: Added compatibility for the default db case.

No matter where  you create hive table the table name must be part of the path and must also be specified here to let sort write to the correct file. The default hive approach is fine. All that's left is to take the default database case into account.

dockerzhang (approved):

woofyzhao:
It turns out that the user should specify full path rather than just warehouse prefix.

Issue #4251, opened 2022-05-18
[Improve][GitHub] Improve GitHub configuration in `.asf.yaml`


shink:
### Description

There are some improvable configuration in `.asf.yaml`, such as `Branch protection`, and so on.
Please check out https://cwiki.apache.org/confluence/exportword?pageId=127405038

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #4257, opened 2022-05-18
[INLONG-4251][GitHub] Improve GitHub configuration in `.asf.yaml`

shink:
fix: #4251

### Motivation

There are some improvable configuration in `.asf.yaml`, such as `Branch protection`, and so on.
Please check out https://cwiki.apache.org/confluence/exportword?pageId=127405038

### Modifications

- add `Branch protection` configuration in `.asf.yaml`

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -68,22 +68,6 @@ github:
 
       required_signatures: false
 
-    # disable force push from the following branches
-    release-0.1.0: {}
-    release-0.2.0: {}
-    release-0.3.0: {}
-    release-0.4.0: {}
-    release-0.5.0: {}
-    release-0.6.0: {}
-    release-0.7.0: {}
-    release-0.8.0: {}
-    release-0.9.0: {}
-    release-0.10.0: {}
-    release-0.11.0: {}
-    release-0.12.0: {}
-    release-1.0.0: {}
-    release-1.1.0: {}
-
 notifications:
   commits:      commits@inlong.apache.org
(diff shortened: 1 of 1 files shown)


dockerzhang (approved):

.asf.yaml
```diff
@@ -46,6 +46,43 @@ github:
...
+    release-0.5.0: {}
+    release-0.6.0: {}
+    release-0.7.0: {}
+    release-0.8.0: {}
+    release-0.9.0: {}
+    release-0.10.0: {}
+    release-0.11.0: {}
+    release-0.12.0: {}
+    release-1.0.0: {}
+    release-1.1.0: {}
```
dockerzhang, line 85: list all release branch is not convenient to maintain

dockerzhang: Can regular matching be used here?

healchow: @shink PTAL, tks.

healchow (changes requested):

.asf.yaml
shink: > list all release branch is not convenient to maintain

@dockerzhang Removed.

shink: > Can regular matching be used here?

There is more info on this on the [PR1678](https://github.com/apache/infrastructure-puppet/pull/1678) that implemented this cool feature. The regular matching is not supported now.

healchow (approved):

PR #4269, opened 2022-05-19
[INLONG-4251][GitHub] Revert the change of .asf.yaml

healchow:
Reverts apache/incubator-inlong#4257


healchow:
I will and this change in other PR.

PR #4323, opened 2022-05-23
[INLONG-4315][Manager] Fix incorrect task service node order in create-group workflow

### LINKED ISSUE #4315: [Bug][Manager] Incorrect task service node order in create-group workflow
### What happened

The sort task service node is before the sink resource node, which seems incorrect. 

The sort task depends on the resources so the resources must be created first. The correct order should be: init_source -> init_mq -> init_sink -> config/start sort

### How to reproduce

Delay the sink resource creation processing time.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
Fixes #4315

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/workflow/group/CreateGroupWorkflowDefinition.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/workflow/group/CreateGroupWorkflowDefinition.java
@@ -20,11 +20,11 @@ package org.apache.inlong.manager.service.workflow.group;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.inlong.manager.common.pojo.workflow.form.GroupResourceProcessForm;
 import org.apache.inlong.manager.service.workflow.ProcessName;
-import org.apache.inlong.manager.service.workflow.listener.GroupTaskListenerFactory;
 import org.apache.inlong.manager.service.workflow.WorkflowDefinition;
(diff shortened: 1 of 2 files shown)


healchow (changes requested):
Please update the order in `CreateStreamWorkflowDefinition` together.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/workflow/group/CreateGroupWorkflowDefinition.java
```diff
@@ -85,14 +85,6 @@ public WorkflowProcess defineProcess() {
         initMQResourceTask.addListenerProvider(groupTaskListenerFactory);
         process.addTask(initMQResourceTask);
 
-        // init Sort resource
```
healchow, line 88: We need to adjust the order of creating resources to 1. MQ -> 2. Sink -> 3. Sort -> 4. Source, so as to ensure that after some links fail, data will not be collected or consumed.

healchow (changes requested):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/workflow/group/CreateGroupWorkflowDefinition.java
woofyzhao: Understood.

healchow (approved):

dockerzhang (approved):

gong (approved):

PR #4450, opened 2022-05-31
[INLONG-4448][Sort] Add Greenplum database data load support

### LINKED ISSUE #4448: [Feature][Sort] Add Greenplum database data load support
### Description

Add Greenplum database data load support, that is to say you can use it to load data to Greenplum database.

### Use case

Use it to load data into Greenplum database all right.

### Code of Conduct

yunqingmoswu:
### Title Name: [INLONG-4448][Sort] Add Greenplum database data load support

Fixes #4448 

### Motivation

I want to support data load to Greenplum database.

### Modifications

It support data load to Greenplum database.

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/GreenplumLoadNode.java
+++ b/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/GreenplumLoadNode.java
@@ -1,0 +1,112 @@
+/*
+ *   Licensed to the Apache Software Foundation (ASF) under one
+ *   or more contributor license agreements.  See the NOTICE file
+ *   distributed with this work for additional information
+ *   regarding copyright ownership.  The ASF licenses this file
+ *   to you under the Apache License, Version 2.0 (the
+ *   "License"); you may not use this file except in compliance
+ *   with the License.  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+
+package org.apache.inlong.sort.protocol.node.load;
+
+import com.google.common.base.Preconditions;
+import lombok.Data;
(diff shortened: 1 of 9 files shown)


GanfengTan (approved):

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/GreenplumLoadNode.java
```diff
@@ -0,0 +1,112 @@
...
+@EqualsAndHashCode(callSuper = true)
+@JsonTypeName("greenplumLoad")
+@Data
+@NoArgsConstructor
+public class GreenplumLoadNode extends LoadNode implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    /**
+     * jdbc:postgresql://host:port/database
```
thexiay, line 51: "pg" change comment here

inlong-sort/sort-connectors/jdbc/src/main/java/org/apache/inlong/sort/jdbc/table/JdbcDialects.java
```diff
@@ -54,13 +58,22 @@ public static Optional<JdbcDialect> get(String url) {
         return Optional.empty();
     }
 
+    public static Optional<JdbcDialect> getCustomDialect(String dialectImpl) {
```
thexiay, line 61: is it better to put `getCustomDialect` wrapped in `get` func . Otherwise user sould know customDialect and normal dialect?

gong (approved):

inlong-sort/sort-connectors/jdbc/src/main/java/org/apache/inlong/sort/jdbc/table/JdbcDialects.java
yunqingmoswu: The user does not need to perceive the difference between the two, but if the user specifies a certain dialect, we must use this dialect, and this effect cannot be achieved only by get, because the url may be the same.

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/GreenplumLoadNode.java
yunqingmoswu: It's not wrong, because the Greenplum also uses "pg".

healchow (approved):

PR #4536, opened 2022-06-06
[INLONG-4535][Agent] Support configurable automatic exit function when OOM happens

### LINKED ISSUE #4535: [Feature][Agent] Support configurable automatic exit function when OOM happens
### Description

_No response_

### Use case

_No response_

### Code of Conduct

pocozh:
### Prepare a Pull Request

- [INLONG-4535][Agent] Support configurable automatic exit function when OOM happens

*(The following *XYZ* should be replaced by the actual [GitHub Issue](https://github.com/apache/incubator-inlong/issues) number)*

- Fixes #4535 

### Motivation
- Generally, the occurrence of OutOfMemoryError means that agent won't run and collect data as ususal. To solve this problem and avoid human manipulation , we let agent supports automatically exit when OOM happens. Therefore, users can use scripts or other detection tools to monitor whether agent exits and restarts it if necessary.

### Modifications
- If "agent.enable.oom.exit" parameter is set as "true" in agent.properties, agent will automatically exit when OutOfMemoryError happens. 

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
+++ b/inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
@@ -26,7 +26,7 @@ import org.slf4j.LoggerFactory;
  */
 public class ThreadUtils {
 
-    private static final Logger logger = LoggerFactory.getLogger(ThreadUtils.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(ThreadUtils.class);
 
     public static void threadThrowableHandler(Thread t, Throwable e) {
         if (AgentUtils.enableOOMExit()) {
@@ -36,7 +36,7 @@ public class ThreadUtils {
 
     private static void handleOOM(Thread t, Throwable e) {
         if (ExceptionUtils.indexOfThrowable(e, java.lang.OutOfMemoryError.class) != -1) {
-            logger.error("Agent exit caused by {} OutOfMemory: ", t.getName(), e);
+            LOGGER.error("Agent exit caused by {} OutOfMemory: ", t.getName(), e);
             forceShutDown();
         }
     }
@@ -45,7 +45,7 @@ public class ThreadUtils {
         try {
             Runtime.getRuntime().exit(-1);
         } catch (Throwable e) {
-            logger.error("exit failed, just halt, exception: ", e);
+            LOGGER.error("exit failed, just halt, exception: ", e);
             Runtime.getRuntime().halt(-2);
         }
     }
--- a/inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
+++ b/inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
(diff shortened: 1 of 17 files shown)


EMsnap (commented):
please add some ut for the oom detection and exit

inlong-agent/agent-core/src/main/java/org/apache/inlong/agent/core/HeartbeatManager.java
```diff
@@ -138,8 +139,9 @@ private Runnable heartBeatReportThread() {
                     int heartbeatInterval = conf.getInt(AGENT_HEARTBEAT_INTERVAL,
                             DEFAULT_AGENT_HEARTBEAT_INTERVAL);
                     TimeUnit.SECONDS.sleep(heartbeatInterval);
-                } catch (Exception ex) {
+                } catch (Throwable ex) {
                     LOGGER.error("error caught", ex);
```
Greedyu, line 143: There is also log processing in the threadThrowableHandler method. Can the log output in the catch later be placed in it?

inlong-agent/conf/agent.properties
```diff
@@ -53,6 +53,7 @@ thread.pool.await.time=30
 agent.local.ip=127.0.0.1
 agent.local.uuid=
 agent.local.uuid.open=false
+agent.enable.oom.exit=false
```
zk1510, line 56: maybe default true?

inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * ThreadUtils, used for handle specified throwable, such as oom, etc.
+ */
+public class ThreadUtils {
+
+    private static final Logger logger = LoggerFactory.getLogger(ThreadUtils.class);
```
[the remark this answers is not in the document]
pocozh: done

inlong-agent/agent-common/src/main/java/org/apache/inlong/agent/utils/ThreadUtils.java
```diff
@@ -0,0 +1,54 @@
...
+package org.apache.inlong.agent.utils;
+
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * ThreadUtils, used for handle specified throwable, such as oom, etc.
+ */
+public class ThreadUtils {
```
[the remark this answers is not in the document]
pocozh: > The function of this class seems to handle exceptions, why is it called ThreadUtils.

Because it is used for catching OOM and exiting thread

inlong-agent/agent-core/src/main/java/org/apache/inlong/agent/core/HeartbeatManager.java
pocozh: > There is also log processing in the threadThrowableHandler method. Can the log output in the catch later be placed in it?

There may be  some other error messages, such as 'LOGGER.error("error while init sender for group id", ex)'.  Moving it  into threadThrowableHandler method does not reflect some key information.

pocozh:
> please add some ut for the oom detection and exit

done

inlong-agent/conf/agent.properties
pocozh: > maybe default true?

At present, there is no relevant detection tool. It is better to temporarily set it as false, otherwise users will not perceive it if the agent exits.

Greedyu (approved):

zk1510 (approved):

healchow (approved):

dockerzhang (approved):

PR #4748, opened 2022-06-23
Delete .idea directory

qiyubin:
it's a not valid file


qiyubin (commented):
.idea files always ignore

healchow:
Hi, thanks for your contribution.
But, this file cannot be deleted, because we need it to add links in IDEA:

image

PR #4749, opened 2022-06-23
[INLONG-4749][Sort] Fix Structure

qiyubin:
./idea is a not valid file


qiyubin:
this file is not necessary upload

healchow:
Hi, thanks for your contribution.
But, this file cannot be deleted, because we need it to add links in IDEA:

image

PR #4882, opened 2022-07-05
[INLONG-4814][Sort][Manager] Declare the primary key for sink table when the query contains update/delete

### LINKED ISSUE #4814: [Bug][Sort][Manager] Declare the primary key for sink table when the query contains update/delete
### What happened

When creating a binlog type data source, an exception occurs when initializing sort: org.apache.flink.client.program.ProgramInvocationException: The main method caused an error: please declare primary key for sink table when query contains update/delete record.
wecom-temp-20a4631da2c50b05cc5e6953c34d0f8c

no error

### How to reproduce

When creating a binlog type data source, an exception occurs when initializing sort.

### Environment

centos7

### InLong version

1.2.0

### InLong Component

InLong Manager, InLong Sort

### Code of Conduct

Greedyu:
Handling exceptions: please declare primary key for sink table when query contains update/delete record.

- Fixes #4814 

Briefly describe ideas

1. The problem is that with flink JDBC SQL you have to specify the primary key.
wecom-temp-5cf30d79023ca92a09bfb0eaee00ac84

2. When the Sort module sees the processing of primaryKey, it should be a problem with the parameters in the node. Gotta go check it out
wecom-temp-ae9621b69bbce4feb44abe2b79454ee2

3. Debug test locates that the primaryKey in the file read by sort is indeed empty
image

4.  To query the source of assignment of GroupInfo.StreamInfo.nodes, there are only CreateSortConfigListenerV2 and CreateStreamSortConfigListener classes, and they are both created by the createNodesForStream() method.
image

5. Test after bug fixes
image

### DIFF
--- a/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
+++ b/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
@@ -37,6 +37,11 @@ import java.io.Serializable;
 import java.util.List;
 import java.util.Map;
 
+/**
+ * clickHouse update operations is heavy, it will takes precedence over all others operation to execute, so it has a
+ * poor performance. It is more suitable for batch scenes but not streaming scenes where operations are performed
+ * frequently.
+ */
 @EqualsAndHashCode(callSuper = true)
 @JsonTypeName("clickHouseLoad")
 @Data
--- a/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
+++ b/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
@@ -61,6 +61,18 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
     @Nonnull
     private String password;
 
+    @JsonProperty("engine")
+    private String engine;
+
+    @JsonProperty("partitionBy")
+    private String partitionBy;
+
+    @JsonProperty("orderBy")
+    private String orderBy;
+
+    @JsonProperty("primaryKey")
+    private String primaryKey;
+
     @JsonCreator
     public ClickHouseLoadNode(@JsonProperty("id") String id,
             @JsonProperty("name") String name,
@@ -73,12 +85,21 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
             @Nonnull @JsonProperty("tableName") String tableName,
             @Nonnull @JsonProperty("url") String url,
             @Nonnull @JsonProperty("userName") String userName,
-            @Nonnull @JsonProperty("passWord") String password) {
+            @Nonnull @JsonProperty("passWord") String password,
+            @JsonProperty("engine") String engine,
+            @JsonProperty("partitionBy") String partitionBy,
+            @JsonProperty("orderBy") String orderBy,
+            @JsonProperty("primaryKey") String primaryKey
+    ) {
         super(id, name, fields, fieldRelations, filters, filterStrategy, sinkParallelism, properties);
         this.tableName = Preconditions.checkNotNull(tableName, "table name is null");
         this.url = Preconditions.checkNotNull(url, "url is null");
         this.userName = Preconditions.checkNotNull(userName, "userName is null");
         this.password = Preconditions.checkNotNull(password, "password is null");
+        this.engine = engine;
+        this.partitionBy = partitionBy;
+        this.orderBy = orderBy;
+        this.primaryKey = primaryKey;
     }
 
     @Override
@@ -100,7 +121,7 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
 
     @Override
     public String getPrimaryKey() {
-        return super.getPrimaryKey();
+        return primaryKey;
     }
 
     @Override
--- a/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
+++ b/inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
@@ -66,15 +66,6 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
     @Nonnull
     private String password;
 
-    @JsonProperty("engine")
-    private String engine;
-
-    @JsonProperty("partitionBy")
-    private String partitionBy;
-
-    @JsonProperty("orderBy")
-    private String orderBy;
-
     @JsonProperty("primaryKey")
     private String primaryKey;
 
@@ -91,9 +82,6 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
             @Nonnull @JsonProperty("url") String url,
             @Nonnull @JsonProperty("userName") String userName,
             @Nonnull @JsonProperty("passWord") String password,
-            @JsonProperty("engine") String engine,
-            @JsonProperty("partitionBy") String partitionBy,
-            @JsonProperty("orderBy") String orderBy,
             @JsonProperty("primaryKey") String primaryKey
     ) {
         super(id, name, fields, fieldRelations, filters, filterStrategy, sinkParallelism, properties);
@@ -101,9 +89,6 @@ public class ClickHouseLoadNode extends LoadNode implements Serializable {
         this.url = Preconditions.checkNotNull(url, "url is null");
         this.userName = Preconditions.checkNotNull(userName, "userName is null");
         this.password = Preconditions.checkNotNull(password, "password is null");
-        this.engine = engine;
-        this.partitionBy = partitionBy;
-        this.orderBy = orderBy;
         this.primaryKey = primaryKey;
     }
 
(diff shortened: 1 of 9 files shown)


thesumery:
There are some reasons that clickhouse sink didn't release the ability of update/delete：
1. clickhouse `update` operations is heavy, it will takes precedence over all others operation to execute, so it has a poor performance. It is more suitable for batch scenes but not streaming scenes where operations are performed frequently.

thesumery:
If user want to use update scenes, how about this two different solution：
1. Import `update` ability，but should tell users the limitation（all code and docs should comment it). In additional what you have submitte in this pr, you should also change update and delete statement in `ClickHouseDialect`.
2. Enable upstream `append-mode`，so upstream produce record is only append stream. Add `time` and `rowkind` field in clickhouse table, those two filed comes from upstream table metadata fileds.Usersdecides how to deduplicate the downstream olap sql.

Greedyu:
> @yunqingmoswu @Greedyu If user want to use update scenes, how about this two different solution：
> 
> 1. Import `update` ability，but should tell users the limitation（all code and docs should comment it). In additional what you have submitte in this pr, you should also change update and delete statement in `ClickHouseDialect`.
> 2. Enable upstream `append-mode`，so upstream produce record is only append stream. Add `time` and `rowkind` field in clickhouse table, those two filed comes from upstream table metadata fileds.Usersdecides how to deduplicate the downstream olap sql.

I thought this was just a bug before, but I didn't expect this ability to be purposely not supported.
But there is still a question, why does the normal mode create group also generate Flink cdc sql

inlong-sort/sort-common/src/test/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNodeTest.java
```diff
@@ -44,6 +44,11 @@ public Node getTestObject() {
                 "ck_demo",
                 "jdbc:clickhouse://localhost:8023/default",
                 "root",
-                "root");
+                "root",
+                "Log",
```
healchow, line 48: The `Log` was one of the engines of ClickHouse?

Greedyu: yes, [https://clickhouse.com/docs/en/engines/table-engines/log-family/log/](https://clickhouse.com/docs/en/engines/table-engines/log-family/log/)

EMsnap (commented):
please add some commentary on the cost of clickhouse update when added primary key

healchow (approved):

inlong-sort/sort-end-to-end-tests/src/test/java/org/apache/inlong/sort/tests/ClickHouseCase.java
```diff
@@ -0,0 +1,161 @@
...
+        if (CLICK_HOUSE_CONTAINER != null) {
+            CLICK_HOUSE_CONTAINER.stop();
+        }
+    }
+
+    private void initializeClickHouseTable() {
+        try {
+            Class.forName(CLICK_HOUSE_CONTAINER.getDriverClassName());
+            Connection conn = DriverManager
+                    .getConnection(CLICK_HOUSE_CONTAINER.getJdbcUrl(), CLICK_HOUSE_CONTAINER.getUsername(),
```
haibo-duan, line 90: please close the JDBC connection At the end of your JDBC program.

Greedyu: resolve

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
```diff
@@ -73,12 +85,21 @@ public ClickHouseLoadNode(@JsonProperty("id") String id,
             @Nonnull @JsonProperty("tableName") String tableName,
             @Nonnull @JsonProperty("url") String url,
             @Nonnull @JsonProperty("userName") String userName,
-            @Nonnull @JsonProperty("passWord") String password) {
+            @Nonnull @JsonProperty("passWord") String password,
+            @JsonProperty("engine") String engine,
+            @JsonProperty("partitionBy") String partitionBy,
```
thesumery, line 90: why add `partitionBy`、`orderBy`、`engine` params?  Where are they used？

Greedyu: These are important attributes of clickhouse. Even if you don't use them directly now, you can add them first.

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
```diff
@@ -100,7 +121,7 @@ public String genTableName() {
 
     @Override
     public String getPrimaryKey() {
-        return super.getPrimaryKey();
+        return primaryKey;
     }
 
     @Override
```
thesumery, line 127: MayBe this is unnecessary.Clickhouse sink currently is implement by jdbc connector.But [jdbc table connector  sink](https://nightlies.apache.org/flink/flink-docs-master/api/java/org/apache/flink/connector/jdbc/table/JdbcDynamicTableSink.html) don't support applyPartition currently. So get partitionFiled is useless.

Greedyu: partitionFiled is useless, the method for obtaining partitionFiled is not modified here, but the getPrimaryKey() method is used

Greedyu:
> added

Add in ClickHouseLoadNode class

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
yunqingmoswu: `partitionBy` can be replaced by `partitionFields` such as `HiveLoadNode`.

thesumery: > why add `partitionBy`、`orderBy`、`engine` params? Where are they used？

ok, allright , although it is useless temporally.

inlong-sort/sort-connectors/jdbc/src/main/java/org/apache/inlong/sort/jdbc/table/JdbcDialects.java
```diff
@@ -44,6 +45,7 @@ public final class JdbcDialects {
         DIALECTS.add(new TDSQLPostgresDialect());
         DIALECTS.add(new SqlServerDialect());
         DIALECTS.add(new OracleDialect());
+        DIALECTS.add(new ClickHouseDialect());
```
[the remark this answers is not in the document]
Greedyu: If it is not added to the list, the unit test will be abnormal here
企业微信截图_f539e39a-2dcc-4498-99d7-bb1d593e032f

inlong-sort/sort-end-to-end-tests/src/test/resources/flinkSql/clickhouse_test.sql
```diff
@@ -0,0 +1,33 @@
...
+CREATE TABLE test_output1 (
+    `id` INT primary key,
+    name STRING,
+    description STRING
+) WITH (
+    'connector' = 'jdbc-inlong',
+    'url' = 'jdbc:clickhouse://clickhouse:8123/default',
+    'table-name' = 'test_output1',
+    'username' = 'default',
+    'password' = ''
```
gong, line 26: Add option "dialect-impl"="org.apache.inlong.sort.jdbc.dialect.ClickHouseDialect". It will can find custom dialect

inlong-sort/sort-connectors/jdbc/src/main/java/org/apache/inlong/sort/jdbc/table/JdbcDialects.java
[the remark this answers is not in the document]
gong: > If it is not added to the list, the unit test will be abnormal here 企业微信截图_f539e39a-2dcc-4498-99d7-bb1d593e032f

UT error reason is  https://github.com/apache/inlong/pull/4882#discussion_r934339275.

inlong-sort/sort-end-to-end-tests/src/test/resources/flinkSql/clickhouse_test.sql
Greedyu: I understand what you mean, but isn't it better to be able to identify the dialect based on the url? It's not an elegant way to test the need to add dialect configuration.

gong: > I understand what you mean, but isn't it better to be able to identify the dialect based on the url? It's not an elegant way to test the need to add dialect configuration.

clickhouse url is ok. posrgres variant(greenplum, gaussdb, tdsqlpostgresql) url will same. So I just suggest keep unify.

Greedyu: ok, it's been resolved

inlong-sort/sort-common/src/main/java/org/apache/inlong/sort/protocol/node/load/ClickHouseLoadNode.java
Greedyu: These properties have been removed

yunqingmoswu (approved):
LGTM

thesumery:
LGTM

PR #5053, opened 2022-07-14
[INLONG-5052][Manager] The workflow that failed to execute still failed, but returned success

### LINKED ISSUE #5052: [Bug][Manager] The workflow that failed to execute still failed, but returned success
### What happened

image
The workflow that failed to execute still failed, but returned success.
The first execution of the workflow fails, and the page display also fails;
For the second time, even if the execution process still fails, the status of this task has changed successfully - it should not succeed;
The success of step 2 leads to the start of the next process - normally, those who fail cannot continue to execute the following process

The failed task should not show success and proceed to the next process
image

### How to reproduce

After the workflow fails to be executed, click re execute

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

fuweng11:
### Prepare a Pull Request

- Fixes #5052

### Motivation

Ensure that the re executed workflow will no longer behave as successful execution and go to the next step in the case of failure

### Modifications

Add  listeners for workflowtask on the interface where you click execute again

### Verifying this change

- [X] This change is a trivial rework/code cleanup without any test coverage.

  *(please describe tests)*

### Documentation

### DIFF
--- a/inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/core/impl/TaskServiceImpl.java
+++ b/inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/core/impl/TaskServiceImpl.java
@@ -68,6 +68,8 @@ public class TaskServiceImpl implements TaskService {
         WorkflowContext context = workflowContextBuilder
                 .buildContextForTask(taskId, WorkflowAction.COMPLETE, remark, operator);
         processorExecutor.executeComplete(context.getActionContext().getTask(), context);
+        ServiceTask serviceTask = (ServiceTask) context.getActionContext().getTask();
+        serviceTask.initListeners(context);
         return context;
     }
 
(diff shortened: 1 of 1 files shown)


kipshi (approved):

inlong-manager/manager-workflow/src/main/java/org/apache/inlong/manager/workflow/core/impl/TaskServiceImpl.java
```diff
@@ -67,6 +68,8 @@ public WorkflowContext transfer(Integer taskId, String remark, List<String> to,
     public WorkflowContext complete(Integer taskId, String remark, String operator) {
         WorkflowContext context = workflowContextBuilder
                 .buildContextForTask(taskId, WorkflowAction.COMPLETE, remark, operator);
+        ServiceTask serviceTask = (ServiceTask) context.getActionContext().getTask();
```
healchow, line 71: Are the results of `context.getActionContext().getTask()` all `ServiceTask`?

fuweng11: no，just one

dockerzhang (approved):

PR #5170, opened 2022-07-21
[INLONG-5169][Sort] Fix race condition issue of HBaseSinkFunction metric collection

### LINKED ISSUE #5169: [Bug][Sort] Metrics of HBaseSinkFunction are not collected accurately
### What happened

* There is a race condition of collecting metrics of `HBaseSinkFunction`

* The metrics are collected in multiple threads without lock, however the metric collection (such as `SimpleCounter`) is not thread-safe

### How to reproduce

* It can be reproduced easily with a high TPS

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Sort

### Code of Conduct

ifndef-SleePy:
### Prepare a Pull Request

- Fixes #5169 

### Motivation

* Fix race condition issue of HBaseSinkFunction metric collection

### Modifications

### Verifying this change

### Documentation

  - If yes, how is the feature documented? (not applicable)

### DIFF
--- a/inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
+++ b/inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
@@ -38,6 +38,7 @@ import org.apache.hadoop.hbase.client.Connection;
 import org.apache.hadoop.hbase.client.ConnectionFactory;
 import org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException;
 import org.apache.inlong.sort.base.metric.SinkMetricData;
+import org.apache.inlong.sort.base.metric.ThreadSafeCounter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -131,10 +132,14 @@ public class HBaseSinkFunction<T> extends RichSinkFunction<T>
                 String groupId = inLongMetricArray[0];
                 String streamId = inLongMetricArray[1];
                 String nodeId = inLongMetricArray[2];
-                sinkMetricData.registerMetricsForDirtyBytes(groupId, streamId, nodeId, DIRTY_BYTES);
-                sinkMetricData.registerMetricsForDirtyRecords(groupId, streamId, nodeId, DIRTY_RECORDS);
-                sinkMetricData.registerMetricsForNumBytesOut(groupId, streamId, nodeId, NUM_BYTES_OUT);
-                sinkMetricData.registerMetricsForNumRecordsOut(groupId, streamId, nodeId, NUM_RECORDS_OUT);
+                sinkMetricData.registerMetricsForDirtyBytes(groupId, streamId, nodeId, DIRTY_BYTES,
+                    new ThreadSafeCounter());
+                sinkMetricData.registerMetricsForDirtyRecords(groupId, streamId, nodeId, DIRTY_RECORDS,
+                    new ThreadSafeCounter());
+                sinkMetricData.registerMetricsForNumBytesOut(groupId, streamId, nodeId, NUM_BYTES_OUT,
+                    new ThreadSafeCounter());
+                sinkMetricData.registerMetricsForNumRecordsOut(groupId, streamId, nodeId, NUM_RECORDS_OUT,
+                    new ThreadSafeCounter());
                 sinkMetricData.registerMetricsForNumBytesOutPerSecond(groupId, streamId, nodeId,
                         NUM_BYTES_OUT_PER_SECOND);
                 sinkMetricData.registerMetricsForNumRecordsOutPerSecond(groupId, streamId, nodeId,
--- a/inlong-sort/sort-connectors/connector-base/src/main/java/org/apache/inlong/sort/base/metric/ThreadSafeCounter.java
+++ b/inlong-sort/sort-connectors/connector-base/src/main/java/org/apache/inlong/sort/base/metric/ThreadSafeCounter.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sort.base.metric;
+
+import org.apache.flink.metrics.Counter;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * A thread-safe implementation of {@link Counter}.
+ */
+public class ThreadSafeCounter implements Counter {
+
+    private final AtomicLong count;
+
+    public ThreadSafeCounter() {
+        count = new AtomicLong(0);
+    }
+
+    @Override
+    public void inc() {
+        count.incrementAndGet();
+    }
+
+    @Override
+    public void inc(long n) {
+        count.addAndGet(n);
+    }
+
+    @Override
+    public void dec() {
+        count.decrementAndGet();
+    }
+
+    @Override
+    public void dec(long n) {
+        count.addAndGet(-n);
+    }
+
+    @Override
+    public long getCount() {
+        return count.get();
+    }
+}
--- a/inlong-sort/sort-connectors/connector-base/src/test/java/org/apache/inlong/sort/base/metric/ThreadSafeCounterTest.java
+++ b/inlong-sort/sort-connectors/connector-base/src/test/java/org/apache/inlong/sort/base/metric/ThreadSafeCounterTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sort.base.metric;
+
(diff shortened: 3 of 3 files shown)


dockerzhang:
@yunqingmoswu @thexiay PTAL, thanks.

inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
```diff
@@ -233,39 +213,20 @@ public void invoke(T value, Context context) throws Exception {
         checkErrorAndRethrow();
 
         mutator.mutate(mutationConverter.convertToMutation(value));
-        rowSize++;
-        dataSize = dataSize + value.toString().getBytes(StandardCharsets.UTF_8).length;
+        if (metricData.getNumRecordsOut() != null) {
+            metricData.getNumRecordsOut().inc(1);
```
EMsnap, line 217: metrics should be updated after flush

inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
```diff
@@ -233,39 +213,20 @@ public void invoke(T value, Context context) throws Exception {
...
-                            .inc(dataSize);
-                }
-                resetStateAfterFlush();
-            } catch (Exception e) {
-                if (metricData.getDirtyRecords() != null) {
-                    metricData.getDirtyRecords().inc(rowSize);
-                }
-                if (metricData.getDirtyBytes() != null) {
-                    metricData.getDirtyBytes().inc(dataSize);
-                }
```
gong, line 257: lost write Fail records

gong:
Maybe we can define a `ConcurrentCounter`

inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
```diff
@@ -80,13 +79,10 @@
     private transient ScheduledExecutorService executor;
     private transient ScheduledFuture scheduledFuture;
     private transient AtomicLong numPendingRequests;
-    private transient RuntimeContext runtimeContext;
 
     private transient volatile boolean closed = false;
 
     private MetricData metricData;
-    private Long dataSize = 0L;
```
yunqingmoswu, line 88: The original logic should be fine, just turn it into an atomic variable.

inlong-sort/sort-connectors/hbase/src/main/java/org/apache/inlong/sort/hbase/sink/HBaseSinkFunction.java
```diff
@@ -80,13 +79,10 @@
...
-    private Long rowSize = 0L;
```
yunqingmoswu, line 89: The original logic should be fine, just turn it into an atomic variable.

yunqingmoswu:
> @yunqingmoswu @thexiay PTAL, thanks.

yunqingmoswu:
> @yunqingmoswu @thexiay PTAL, thanks.

done

ifndef-SleePy:
Hi guys, thanks for reviewing. First, I'm sorry that I didn't explain my points behind the PR. 

As you can see, all the comments are focused on the scenario of flushing failure. I think we don't need care about the metrics after flushing failure. Because
1. If flushing fails, the job would falls into failover scenario.
2. Metric accuracy can not be guaranteed when failover happens. The consistency mode of `HBaseSinkFunction` is at-least-once. So some of produced records would probably be re-produced again. However the metric collection would not be reverted. All of these records would be counted.
3. Moreover, nobody knows how many records have been written into HBase Region Server when flushing fails. Flushing of `BufferedMutator` is not a transactional operation. If I remember correctly, HBase client spawns RPC calls to several Region Server concurrently. Any failure RPC call leads to a global failure. So when failure happens, we don't know how many RPC succeeded. It might be none, some or all records have been flushed successfully.

My points here are:
1. We have to guarantee the metric accuracy if no failure happens. That's what we do in this PR.
2. Metric is not accurate if flushing fails, no matter what we do. So we don't need to involve any synchronization of these counters. Just make it as cheaper as possible.
3. A counter records the number of flushing failure times is meaningful. But I'm a bit confused of the counter name "dirty", so I removed the these ambiguous counters. Maybe we could introduce another counter to record the failure times.

What do you guys think of this? @EMsnap @gong @yunqingmoswu

yunqingmoswu:
> Hi guys, thanks for reviewing. First, I'm sorry that I didn't explain my points behind the PR.
> 
> As you can see, all the comments are focused on the scenario of flushing failure. I think we don't need care about the metrics after flushing failure. Because
> 
> 1. If flushing fails, the job would falls into failover scenario.
> 2. Metric accuracy can not be guaranteed when failover happens. The consistency mode of `HBaseSinkFunction` is at-least-once. So some of produced records would probably be re-produced again. However the metric collection would not be reverted. All of these records would be counted.
> 3. Moreover, nobody knows how many records have been written into HBase Region Server when flushing fails. Flushing of `BufferedMutator` is not a transactional operation. If I remember correctly, HBase client spawns RPC calls to several Region Server concurrently. Any failure RPC call leads to a global failure. So when failure happens, we don't know how many RPC succeeded. It might be none, some or all records have been flushed successfully.
> 
> My points here are:
> 
> 1. We have to guarantee the metric accuracy if no failure happens. That's what we do in this PR.
> 2. Metric is not accurate if flushing fails, no matter what we do. So we don't need to involve any synchronization of these counters. Just make it as cheaper as possible.
> 3. A counter records the number of flushing failure times is meaningful. But I'm a bit confused of the counter name "dirty", so I removed the these ambiguous counters. Maybe we could introduce another counter to record the failure times.
> 
> What do you guys think of this? @EMsnap @gong @yunqingmoswu

As far as I know, the underlying logic of refresh should generate hfile first and then load, then refresh will only have this batch of data either all visible or all lost. If I understand correctly, then after refreshing the statistics, the data obtained will be more accurate, not only for dirty data, but also for normal sync data.

gong:
> 

@ifndef-SleePy hello, here define of dirty data is write fail, and we need data records and data size.

ifndef-SleePy:
> > Hi guys, thanks for reviewing. First, I'm sorry that I didn't explain my points behind the PR.
> > As you can see, all the comments are focused on the scenario of flushing failure. I think we don't need care about the metrics after flushing failure. Because
> > 
> > 1. If flushing fails, the job would falls into failover scenario.
> > 2. Metric accuracy can not be guaranteed when failover happens. The consistency mode of `HBaseSinkFunction` is at-least-once. So some of produced records would probably be re-produced again. However the metric collection would not be reverted. All of these records would be counted.
> > 3. Moreover, nobody knows how many records have been written into HBase Region Server when flushing fails. Flushing of `BufferedMutator` is not a transactional operation. If I remember correctly, HBase client spawns RPC calls to several Region Server concurrently. Any failure RPC call leads to a global failure. So when failure happens, we don't know how many RPC succeeded. It might be none, some or all records have been flushed successfully.
> > 
> > My points here are:
> > 
> > 1. We have to guarantee the metric accuracy if no failure happens. That's what we do in this PR.
> > 2. Metric is not accurate if flushing fails, no matter what we do. So we don't need to involve any synchronization of these counters. Just make it as cheaper as possible.
> > 3. A counter records the number of flushing failure times is meaningful. But I'm a bit confused of the counter name "dirty", so I removed the these ambiguous counters. Maybe we could introduce another counter to record the failure times.
> > 
> > What do you guys think of this? @EMsnap @gong @yunqingmoswu
> 
> As far as I know, the underlying logic of refresh should generate hfile first and then load, then refresh will only have this batch of data either all visible or all lost. If I understand correctly, then after refreshing the statistics, the data obtained will be more accurate, not only for dirty data, but also for normal sync data.

1. The `HFile` way you described is called "bulk loading" [1]. That's not the way worked here. `BufferedMutator` would not trigger bulk loading, there is another API for bulk loading. It would go through WAL and `MemStore` to be ingested into Region Server.
2. Bulk loading also could not guarantee "this batch of data either all visible or all lost", because there might be several `HFile` involved in the batch operation.
3. HBase could guarantee consistency in a row, but not cross rows, please check the ACID docs [2], "APIs that mutate several rows will _not_ be atomic across the multiple rows.".

[1] https://hbase.apache.org/book.html#arch.bulk.load
[2] https://hbase.apache.org/acid-semantics.html

yunqingmoswu (approved):
It is possible to report indicators before and after flushing, but it is recommended to report indicators after flushing.

yunqingmoswu:
> > > Hi guys, thanks for reviewing. First, I'm sorry that I didn't explain my points behind the PR.
> > > As you can see, all the comments are focused on the scenario of flushing failure. I think we don't need care about the metrics after flushing failure. Because
> > > 
> > > 1. If flushing fails, the job would falls into failover scenario.
> > > 2. Metric accuracy can not be guaranteed when failover happens. The consistency mode of `HBaseSinkFunction` is at-least-once. So some of produced records would probably be re-produced again. However the metric collection would not be reverted. All of these records would be counted.
> > > 3. Moreover, nobody knows how many records have been written into HBase Region Server when flushing fails. Flushing of `BufferedMutator` is not a transactional operation. If I remember correctly, HBase client spawns RPC calls to several Region Server concurrently. Any failure RPC call leads to a global failure. So when failure happens, we don't know how many RPC succeeded. It might be none, some or all records have been flushed successfully.
> > > 
> > > My points here are:
> > > 
> > > 1. We have to guarantee the metric accuracy if no failure happens. That's what we do in this PR.
> > > 2. Metric is not accurate if flushing fails, no matter what we do. So we don't need to involve any synchronization of these counters. Just make it as cheaper as possible.
> > > 3. A counter records the number of flushing failure times is meaningful. But I'm a bit confused of the counter name "dirty", so I removed the these ambiguous counters. Maybe we could introduce another counter to record the failure times.
> > > 
> > > What do you guys think of this? @EMsnap @gong @yunqingmoswu
> > 
> > 
> > As far as I know, the underlying logic of refresh should generate hfile first and then load, then refresh will only have this batch of data either all visible or all lost. If I understand correctly, then after refreshing the statistics, the data obtained will be more accurate, not only for dirty data, but also for normal sync data.
> 
> 1. The `HFile` way you described is called "bulk loading" [1]. That's not the way worked here. `BufferedMutator` would not trigger bulk loading, there is another API for bulk loading. It would go through WAL and `MemStore` to be ingested into Region Server.
> 2. Bulk loading also could not guarantee "this batch of data either all visible or all lost", because there might be several `HFile` involved in the batch operation.
> 3. HBase could guarantee consistency in a row, but not cross rows, please check the ACID docs [2], "APIs that mutate several rows will _not_ be atomic across the multiple rows.".
> 
> [1] https://hbase.apache.org/book.html#arch.bulk.load [2] https://hbase.apache.org/acid-semantics.html

Thanks for clarifying.

ifndef-SleePy:
After an offline communication with @yunqingmoswu , we reached an agreement that just keep the origin implementation with thread-safe counter. Because it's bases on production requirement, not technical.
1. We have to keep the meaningless "dirty" prefix counter till someday the production manager realizes that it's meaningless.
2. Summing the counter after flushing is better than before flushing although neither of them could guarantee accuracy. Because we don't want to meet the scenario that user asks a question why there exists successful counter in metric system however there is no data in HBase.

gong:
please git pull uptream master , git rebase , resolve conflic@ifndef-Sleepy

ifndef-SleePy:
I've re-created the PR after resolving conflicts.

gong (approved):

thesumery (approved):

healchow (approved):

yunqingmoswu (approved):
LGTM

EMsnap (approved):

PR #5173, opened 2022-07-21
[INLONG-5167][Manager][Agent][Dataproxy][SDK] Manager open API authentication support

### LINKED ISSUE #5167: [Improve][Manager] OpenAPI authentication support
### Description

Currently the manager's open APIs are free to call from anywhere anonymously, which is not secure.
Goal: Add some sort of authentication support to very the identity of the clients (i.e. authorized agent/dataproxy/sdk etc) 

### Mechanisms
There are several ways to provide web api authentication, with the typical two described below:

**1. Basic Access Authentication**
https://en.wikipedia.org/wiki/Basic_access_authentication

image

The auth process is simple:
A) The api server side generates a secretID  & secretKey pair for each client who wants access.
B) The client encode the id & key pair (base64) in the http header.
C) The server's interceptor / filter retrieves the secretID and compares the secretKey param with the true secretKey stored at some database (usually encrypted)

**Pros:**  
  - Light and easy to understand
  - little extra work to use

**Cons:**  
  - Since base64 is used, it is not secure in non-encrypted channel. 
  - So it **must be used along with HTTPS.**

**2. API Secure Signature**
This is commonly used in public cloud apis, like the one described in this [Tencent cloud approach](https://cloud.tencent.com/document/product/213/30654).

image

The auth process is roughly like this:
A) The api server side generates a secretID  & secretKey pair for each client who want access.
B) The client organize some of the request context info (URI、domain、sorted params、body data etc) into a basic string. 
C) Then client use RSA or hashing to encrypt this string to a signature with the secretKey
D) The client request the API with the signature and secretID
E) The server retrieves the secretID, find its corresponding secretKey, and uses the same encoding process as B) and C) to calculate the signature.
F) The server compares its calculated signature and the one from the request to determine the final authenticating result.

**Pros:**
  - More secure, since the secretKey is not transported across the network but merely as part of the signature.  Thus can also be used in unencrypted channels like HTTP. 

**Cons:**
  - The process is more sophisticated and thus error prone.  Steps like B) has no standards to follow. Often need more debugging time for the user.

**Suggest choosing the basic auth approach** for the following reasons:
1) We should always use HTTPs as a modern web application. Thus we enforce it and basic auth is easy and sufficiently secure.
2) The open APIs exposed by manager are quite limited and are usually read only (Most important of which includes configuration query for agency and dataproxy

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
## Prepare a Pull Request

- Fixes #5167

## Motivation

The Manager's Open APIs are free to call from anywhere anonymously, which is not secure.

Some sort of authentication support is needed to very the identity of the clients (i.e. authorized Agent/DataProxy/SDK, etc.).

## Changes

- Open API basic auth integration with Apache Shiro.
- Client component basic auth header added.
- Default secretId & secretKey pair configuration to keep the suite ready to go out of the box.
 (User should replace with their own if they care)

## Mechanism Options

There are several ways to provide web API authentication, with the typical two described below:

### Approach 1: Basic Access Authentication

See at: https://en.wikipedia.org/wiki/Basic_access_authentication  or [rfc7617](https://www.rfc-editor.org/rfc/rfc7617)

image

The auth process is simple:

A) The API server side creates a secretID & secretKey pair for each client who wants access.
B) The client encodes the id & key pair (base64) in the HTTP header.
C) The server's interceptor/filter retrieves the secretID and compares the secretKey param with the true secretKey stored at some database (usually encrypted)

**Pros:**  
  - Light and easy to understand, secure enough with HTTPS
  - little extra work to use

**Cons:**  
  - Since base64 is used, it is not secure in the non-encrypted channels.  The encoded credential string is always the same.
  - So it **must be used along with HTTPS.**

### Approach 2: API Secure Signature

This is commonly used in public cloud APIs, like the one described in this [Tencent cloud approach](https://cloud.tencent.com/document/product/213/30654)  or [rfc7616](https://www.rfc-editor.org/rfc/rfc7616)

image

The auth process is roughly like this:
A) The API server side creates a secretID  & secretKey pair for each client who wants access.
B) The client organizes the request context info (URI、domain、sorted params、body data etc) into a concatenated string. 
C) Then client use RSA or hashing to encrypt the concatenated string to a signature with the secretKey
D) The client requests the API with the signature and secretID
E) The server retrieves the secretID, finds its corresponding secretKey, and uses the same signing process as B) and C) to calculate again the signature.
F) The server compares its calculated signature and the one from the request to determine the final authenticating result.

**Pros:**
  - More secure, since **the secretKey is not transported across the network** but merely as part of the signature source.  Thus can also be used in unencrypted channels like HTTP. 
  - The signature changes every time with different request contexts.

**Cons:**
  - The concatenating and signing process is more sophisticated and cumbersome and thus error-prone.  
  - Concatenating process like B) has no standards to follow. Often more debugging time is needed for the end user.

### Other approaches
Other popular methods like the bearer token used in OAuth2.0 ([rfc6750](https://www.rfc-editor.org/rfc/rfc6750)) behaves somewhere between the above two:  the token is different with each request / session but still have the same constraint (should only be used in https).

## Design

**This PR chooses the basic auth approach (Approach 1)** for the following reasons:
1) We must always use HTTPS like every modern web applications. And we leverage it by using basic auth which is easy and sufficiently secure.
2) The open APIs exposed by the Manager are quite limited and usually read only (Most important of which includes configuration queries for Agent and DataProxy).

## Implementation

- Server:  integration with Apache Shiro by using different realms for web user auth and open API auth
- Client: HTTP standard authorization header

### DIFF
--- a/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/openapi/OpenAPIFilter.java
+++ b/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/openapi/OpenAPIFilter.java
@@ -15,9 +15,16 @@
  * limitations under the License.
  */
 
-package org.apache.inlong.manager.web.auth;
+package org.apache.inlong.manager.web.auth.openapi;
 
-import java.io.IOException;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang.StringUtils;
+import org.apache.shiro.SecurityUtils;
+import org.apache.shiro.subject.Subject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
 import javax.servlet.FilterConfig;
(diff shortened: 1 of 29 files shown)


ciscozhou (approved):
LGTM

vernedeng (approved):

showingdata:
I think the CommonConstants class HTTP_AUTH_HEADER name "Authorization" should be changed from lowercase to "Authorization"

healchow:
> I think the CommonConstants class HTTP_AUTH_HEADER name "Authorization" should be changed from lowercase to "authorization"

Hi, the "Authorization" is a more formal way, see more here: https://en.wikipedia.org/wiki/Basic_access_authentication 

image

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/openapi/OpenAPIFilter.java
```diff
@@ -0,0 +1,122 @@
...
+import java.io.IOException;
+import java.util.Base64;
+
+/**
+ * Filter of open api authentication.
+ */
+@Slf4j
+public class OpenAPIFilter implements Filter {
+
+    public static final String BASIC_AUTH_HEADER = "Authorization";
```
healchow, line 45: Suggested moving those constants into a constant class.

woofyzhao:
> I think the CommonConstants class HTTP_AUTH_HEADER name "Authorization" should be changed from lowercase to "authorization"

Both are PK. The http headers are case insensitive

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/openapi/OpenAPIFilter.java
woofyzhao: Done

healchow (approved):
LGTM

dockerzhang:
@woofyzhao there are conflicting files. Please fix it, thanks.

kipshi (approved):

woofyzhao:
> @woofyzhao there are conflicting files. Please fix it, thanks.

done

dockerzhang (approved):

PR #5214, opened 2022-07-26
[INLONG-5097][TubeMQ] Keep the protocol compatible with previous versions

### LINKED ISSUE #5097: [Bug][TubeMQ] The protocol from 1.1.0 and later is not smoothly compatible with previous versions
### What happened

The TubeMQ module protocol from 1.1.0 and later is not smoothly compatible with previous versions

When I tested it, I found that since version 1.1.0, the client and server of TubeMQ cannot communicate with the old version of TubeMQ. I checked the modification history, and the problem is to upgrade TubeMQ in version 1.1.0 [1] Introduced when a module's package depends.

This is a relatively serious problem. We should upgrade the modules smoothly in a compatible form. Even if we want to upgrade the protocol in an incompatible form, we need to make the transition when upgrading the large version number, not in the small version. And only upgrading components to make protocol changes does not introduce new features. This is a serious bug.

@baomingyu Can you take a look at this problem and see how to fix it

1. https://github.com/apache/inlong/issues/3644

Compatible with old and new versions

### How to reproduce

Use the 1.1.0 version to test the interoperability with the previous version

### Environment

_No response_

### InLong version

1.1.0

### InLong Component

InLong TubeMQ

### Code of Conduct

baomingyu:
Fix the protocol from 1.1.0 and later is not smoothly compatible with previous versions

- Fixes #5097

### Motivation

### Modifications

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corerpc/RpcConstants.java
+++ b/inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corerpc/RpcConstants.java
@@ -43,7 +43,6 @@ public final class RpcConstants {
     public static final String NETTY_WRITE_HIGH_MARK = "rpc.netty.write.highmark";
     public static final String NETTY_WRITE_LOW_MARK = "rpc.netty.write.lowmark";
     public static final String NETTY_TCP_SENDBUF = "rpc.netty.send.buffer";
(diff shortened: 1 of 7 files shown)


gosonzhang (approved):
Perfect!

healchow (approved):

PR #5284, opened 2022-07-29
[INLONG-5283][Manager] Add OpenAPI authentication option (default off)

### LINKED ISSUE #5283: [Improve][Manager] OpenAPI authentication should not be mandatory
### Description

The open api authentication better have an option to be turned on or off.
For example when the manager and its clients (agent etc.) are all in the same intranet, there might be no need to do authentication.
Also the default secret key is not safe to be kept in config, so we decided to turn it off be default. User who needs to do open api authentication should configure the secretID and secretKey manually.
The related documentation will follow up soon.

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
- Fixes #5283

See the link above for motivation.

### DIFF
--- a/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/impl/InlongShiroImpl.java
+++ b/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/impl/InlongShiroImpl.java
@@ -60,7 +60,7 @@ public class InlongShiroImpl implements InlongShiro {
     @Autowired
     private UserService userService;
 
-    @Value("${openapi.auth.enabled}")
+    @Value("${openapi.auth.enabled:false}")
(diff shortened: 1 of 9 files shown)


dockerzhang (approved):

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/auth/impl/InlongShiroImpl.java
```diff
@@ -59,6 +60,9 @@ public class InlongShiroImpl implements InlongShiro {
     @Autowired
     private UserService userService;
 
+    @Value("${openapi.auth.enabled}")
```
healchow, line 63: Suggest adding a default value in the `@Value` annotation.

healchow (approved):

Issue #5231, opened 2022-07-27
[Improve][CodeStyle] Add spotless to CI


jun0315:
### Description

Currently, we do not check the code style, so we can add spotless to Ci to check.

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #5379, opened 2022-08-05
[INLONG-5231][Improve][CodeStyle] Add spotless to CI

jun0315:
see https://github.com/apache/inlong/issues/5231

At present, the AOSP format specification is selected for spotless, which should be consistent with the existing format as much as possible. Currently, only spotless apply the manager module.


jun0315:
checkstyle has a lot to adapt, and I'm still adapting...

pom.xml
```diff
@@ -250,6 +251,8 @@
         <HikariCP.version>4.0.3</HikariCP.version>
         <caffeine.version>2.9.3</caffeine.version>
         <kafka.clients.version>3.0.0</kafka.clients.version>
+        <!-- JDK1.8 only support google java format 1.7-->
+        <google.java.format.version>1.7</google.java.format.version>
```
leosanqing, line 255: I can't find a version that says that jdk8 only supports google-java-format1.7 in the official documentation, can you provide the link?  thx

PR #5544, opened 2022-08-15
[INLONG-5101][DataProxy] Optimize load balancing for DataProxy

### LINKED ISSUE #5101: [Feature] Optimize load balancing for DataProxy
### Description

When the SDK produces data, it will select a group of nodes to send data, and when some of them have problems, they will be eliminated, and new candidate nodes will be selected;
It is expected to reduce the impact on production so that the client is unaware; b, after the server node is restored, it can be selected again; c, ensure the node load balance.

### Use case

_No response_

### Code of Conduct

rhizoma-atractylodis:
fix #5101

### Motivation
Based on my reading of the source code. Currently, the DataProxy SDK side selects DataProxy nodes using polling (sending messages in TCP mode) and random selection (sending messages in HTTP mode).The polling method is not efficient enough, and the random method is not easy to achieve load balancing.

### Changes
Use consistent hashing algorithm instead of the original polling and random

### Mechanism Options
Consistent Hash Algorithm and Virtual Node Mechanism
[Refer to the article for details on the algorithm](https://blog.csdn.net/gonghaiyu/article/details/108375298)

### Design
Based on my reading of the source code.The following are the functions that need to be modified:

org.apache.inlong.sdk.dataproxy.network.ClientMgr.getClientByRoundRobin():This function obtains the DataProxy node by polling
org.apache.inlong.sdk.dataproxy.http.InternalHttpSender.sendMessageWithHostInfo(List bodies, String groupId, String streamId, long dt, long timeout, TimeUnit timeUnit):This function implements the selection of DataProxy nodes by randomly selecting HostInfo
Need to update the fields of the DataProxy node class to add information about virtual nodes
The hash ring and virtual nodes need to be completed on the DataProxy side, and the strategy for acquiring DataProxy nodes on the SDK side must be updated at the same time.

### DIFF
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/enums/FieldType.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/enums/FieldType.java
@@ -48,7 +48,8 @@ public enum FieldType {
     LOCAL_ZONE_TIMESTAMP,
     ARRAY,
     MAP,
-    STRUCT;
+    STRUCT,
+    FUNCTION;
 
     public static FieldType forName(String name) {
         Preconditions.checkNotNull(name, "FieldType should not be null");
--- a/inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/sort/util/FieldRelationUtils.java
+++ b/inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/sort/util/FieldRelationUtils.java
@@ -199,7 +199,7 @@ public class FieldRelationUtils {
      */
     private static List<FieldRelation> createEncryptFieldRelations(List<StreamField> fieldList, String transformName,
             EncryptDefinition encryptDefinition, String preNodes, Map<String, StreamField> constantFieldMap) {
-        Preconditions.checkNotEmpty(preNodes, "PreNodes of splitter should not be null");
+        Preconditions.checkNotEmpty(preNodes, "PreNodes of encrypt should not be null");
         String preNode = preNodes.split(",")[0];
         List<EncryptRule> encryptRules = encryptDefinition.getEncryptRules();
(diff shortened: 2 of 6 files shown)


inlong-sdk/dataproxy-sdk/src/main/java/org/apache/inlong/sdk/dataproxy/network/Sender.java
```diff
@@ -513,15 +573,97 @@ private boolean validAttribute(String attr) {
     /**
      * Following methods used by asynchronously message sending.
      */
-    public void asyncSendMessage(EncodeObject encodeObject, SendMessageCallback callback, String msgUUID,
-            long timeout, TimeUnit timeUnit) throws ProxysdkException {
+    public void asyncSendMessage(EncodeObject encodeObject, SendMessageCallback callback, String msgUUID, long timeout,
```
baomingyu, line 584: Why not reuse next asyncSendMessage by specifying a default loadBalance type

inlong-sdk/dataproxy-sdk/src/main/java/org/apache/inlong/sdk/dataproxy/utils/LoadBalance.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sdk.dataproxy.utils;
+
+public class LoadBalance {
```
baomingyu, line 20: Please change it to enum type here

baomingyu (changes requested):

PR #5792, opened 2022-09-06
[INLONG-5788][DataProxy] Fix the incorrect readFailSize metric value

### LINKED ISSUE #5788: [Bug][DataProxy] Incorrect readFailSize metric value
### What happened

In the case that the sent messages are all successful, the readFailSize value is not 0:
image

From the implementation point of view, it should be caused by an implementation bug:
image

So, we need to adjust it.

Should return to zero when all are successful

### How to reproduce

Send success messages and check the readFailSize value

### Environment

Linux 

### InLong version

master

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #5788 
-

### DIFF
--- a/inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/TubeSink.java
+++ b/inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/TubeSink.java
@@ -285,7 +285,7 @@ public class TubeSink extends AbstractSink implements Configurable {
                     cachedMsgCnt.incrementAndGet();
                     DataProxyMetricItem metricItem = this.metricItemSet.findMetricItem(dimensions);
                     metricItem.readSuccessCount.incrementAndGet();
-                    metricItem.readFailSize.addAndGet(event.getBody().length);
+                    metricItem.readSuccessSize.addAndGet(event.getBody().length);
                 } else {
                     tx.rollback();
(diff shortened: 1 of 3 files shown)


luchunliang (approved):
OK

dockerzhang (approved):

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/TubeSink.java
```diff
@@ -285,7 +285,7 @@ public Status process() throws EventDeliveryException {
                     cachedMsgCnt.incrementAndGet();
                     DataProxyMetricItem metricItem = this.metricItemSet.findMetricItem(dimensions);
                     metricItem.readSuccessCount.incrementAndGet();
-                    metricItem.readFailSize.addAndGet(event.getBody().length);
+                    metricItem.readSuccessSize.addAndGet(event.getBody().length);
                 } else {
                     tx.rollback();
                     //logger.info("[{}] Channel --> Queue(has no enough space,current code point) "
```
healchow, line 291: Could you please remove those unused log codes at the same time?

gosonzhang: I see a lot of such useless comments and codes. It is better to submit a separate issue to solve it.
@healchow

healchow (approved):

PR #5837, opened 2022-09-08
[INLONG-5836][Manager] Fix the source status related behavior problems

### LINKED ISSUE #5836: [Bug][Manager] Stream source status related behavior problem fix
### What happened

1. The AUTO_PUSH source type should always be normal.
2. The template source should also be deleted when group is deleted.

correct behavior

### How to reproduce

Delete the group

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
- Fixes #5836

- AUTO_PUSH source type should always be normal.
- Template sources should also be deleted when their groups are deleted.
- Other bugs.


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/source/AbstractSourceOperator.java
```diff
@@ -155,22 +153,25 @@ public void updateOpt(SourceRequest request, Integer groupStatus, String operato
...
-                case SOURCE_NORMAL:
-                    entity.setStatus(SourceStatus.TO_BE_ISSUED_ADD.getCode());
-                    break;
-                case SOURCE_FAILED:
-                    entity.setStatus(SourceStatus.SOURCE_NEW.getCode());
-                    break;
-                default:
-                    // others leave it be
-                    break;
+        if (!entity.getSourceType().equals(SourceType.AUTO_PUSH)) {
```
healchow, line 156: Suggest moving this check to the beginning of this method, if it is `AUTO_PUSH` source, just print a log and then return.

GanfengTan (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
```diff
@@ -215,15 +216,15 @@ private List<DataConfig> fetchFileTasks(TaskRequest taskRequest) {
         Preconditions.checkTrue(StringUtils.isNotBlank(agentIp) || StringUtils.isNotBlank(agentClusterName),
                 "both agent ip and cluster name are blank when fetching file task");
         List<StreamSourceEntity> sourceEntities = sourceMapper.selectByAgentIpOrCluster(needAddStatusList,
-                Lists.newArrayList(SourceType.FILE), agentIp, agentClusterName,TASK_FETCH_SIZE * 10);
+                Lists.newArrayList(SourceType.FILE), agentIp, agentClusterName,UNLIMITED_FETCH_SIZE);
```
healchow, line 219: Suggested removing the limit in SQL.

healchow (changes requested):

dockerzhang (approved):

healchow (approved):

PR #5898, opened 2022-09-15
[INLONG-5897][Manager] Support query audit data by user role

### LINKED ISSUE #5897: [Feature][Manager] Support query audit data by user role
### Description
In some cases:
Only admin needs the complete audit data view.
Ordinary customer user needs only to know data collection count and sort output out.

### Use case

_No response_

### Code of Conduct

woofyzhao:
- Fixes #5897

### Motivation
1.  In some cases, it **may** be needed when admin and ordinary user may need different audit data view, e.g : 
only admin needs the complete audit data view, and ordinary customer user needs only to know data collection count and sort output out.

**The advantage is that customer view can be properly simplified and easier to understand**

2.  When the data source type is AUTO_PUSH there is no agent audit data. Use data-proxy receiving audit data instead.
Similar case for when there is no sort process.

### Modification
- Support audit id configuration by ROLE.
-  Adjust audit id by stream source and sink configuration type.
- Add sink node type in response to give more display info.


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AuditServiceImpl.java
```diff
@@ -128,6 +148,24 @@ public List<AuditVO> listByCondition(AuditRequest request) throws IOException {
         return aggregateByTimeDim(result, request.getTimeStaticsDim());
     }
 
+    @Override
+    public List<AuditVO> listByRole(AuditRequest request) throws IOException {
+        UserInfo userInfo = LoginUserUtils.getLoginUser();
+        List<String> adminAuditIds = Arrays.asList(
```
healchow, line 154: Suggest extracting to an immutable constant.

healchow (approved):

dockerzhang (approved):

yunqingmoswu (commented):
1.Explain why users only pay attention to AUDIT_ID_DATAPROXY_RECEIVED and AUDIT_ID_SORT_OUTPUT?
2.It is suggested that querying audit views based on roles is only a default behavior. Users can also query any audit views according to their own needs.

woofyzhao:
> 1.Explain why users only pay attention to AUDIT_ID_DATAPROXY_RECEIVED and AUDIT_ID_SORT_OUTPUT? 2.It is suggested that querying audit views based on roles is only a default behavior. Users can also query any audit views according to their own needs.

Excellent idea. Fixed as required.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AuditServiceImpl.java
```diff
@@ -70,12 +78,21 @@ public class AuditServiceImpl implements AuditService {
     private static final String HOUR_FORMAT = "yyyy-MM-dd HH";
     private static final String DAY_FORMAT = "yyyy-MM-dd";
 
+    @Value("#{'${audit.admin.ids:3,4,5,6,7,8}'.split(',')}")
```
healchow, line 81: 1. The number is not clear for users.
2. How to set those ids? In a properties file?

woofyzhao: 1. doc and inlong audit definition url link added.
2. set those ids in application.properties like:
audit.admin.ids = 3,4,5,6,7,8
audit.user.ids = 3,8

healchow (approved):

PR #6010, opened 2022-09-23
[INLONG-6003][Manager] Change the transform source node name to be MQ node in standard mode

### LINKED ISSUE #6003: [Bug][Manager] The transform source node relation is incorrect under standard mode
### What happened

The transform input node is wrong

Should be mq node instead of upstream source in sort configuration

### How to reproduce

Use transform under standard mode

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

woofyzhao:
- Fixes #6003

### Motivation

In the beginning only **lightweight mode** supports the ETL transform nodes. Clients configure their extract/transform/load nodes  at will and what they submit is what they see so everyone is happy.

This ceases to be the case when **standard sort mode** tries to support transform nodes as well. In standard mode multiple upstream sources are collected to MQ first and are not the real extract nodes anymore. Instead the MQ cache node becomes **the real extract node**.  

This is achieved by manager through replacing the original upstream source nodes (file、binlog...) with the MQ cache node when constructing the extract node in standard mode.

However the clients are ignorant of any of this.  From their point of view nothing is changed and the MQ node is invisible to them. So in their submitted **node relation configuration** the extract node is still some upstream source node (file、binlog...) rather than the MQ cache node. This results in the incorrect DataFlow configuration:

image

### Changes
- change the transform fields's original node id to be MQ cache node
-  update the user submitted node relation to use MQ cache node as the real extract node

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
@@ -127,16 +127,16 @@ public class DefaultSortConfigOperator implements SortConfigOperator {
                 relations = NodeRelationUtils.createNodeRelations(sources, sinks);
             } else {
                 relations = NodeRelationUtils.createNodeRelations(inlongStream);
-            }
 
-            // redirect transform input fields node if necessary
-            preprocessTransformList(groupInfo, sources, transformResponseList);
+                // redirect transform input fields node if necessary
+                preprocessTransformList(groupInfo, sources, transformResponseList);
+
+                // replace upstream source node id with mq node in standard mode
+                adjustNodeRelations(relations, groupInfo, sources, transformResponseList);
+            }
 
             // create extract-transform-load nodes
             List<Node> nodes = this.createNodes(sources, transformResponseList, sinks, fieldMap);
-
-            // replace upstream source node id with mq node in standard mode
-            adjustNodeRelations(relations, groupInfo, sources, transformResponseList);
 
             StreamInfo streamInfo = new StreamInfo(streamId, nodes, relations);
             sortStreamInfos.add(streamInfo);
@@ -150,12 +150,13 @@ public class DefaultSortConfigOperator implements SortConfigOperator {
 
     private Set<String> getValidInputNodeNamesInStandardMode(List<StreamSource> mqSources,
             List<TransformResponse> transforms) {
-        Set<String> mqSourceNameSet = mqSources.stream().map(StreamSource::getSourceName).collect(Collectors.toSet());
-        Set<String> transformNameSet = transforms.stream().map(TransformResponse::getTransformName)
-                .collect(Collectors.toSet());
         Set<String> result = Sets.newHashSet();
-        result.addAll(mqSourceNameSet);
-        result.addAll(transformNameSet);
+        if (CollectionUtils.isNotEmpty(mqSources)) {
+            result.addAll(mqSources.stream().map(StreamSource::getSourceName).collect(Collectors.toSet()));
+        }
+        if (CollectionUtils.isNotEmpty(transforms)) {
+            result.addAll(transforms.stream().map(TransformResponse::getTransformName).collect(Collectors.toSet()));
+        }
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
```diff
@@ -137,6 +148,58 @@ private GroupInfo getGroupInfo(InlongGroupInfo groupInfo, List<InlongStreamInfo>
...
+        }
+
+        // set relations' input node to mq node when necessary
+        String mqNodeName = sources.get(0).getSourceName();
+        Set<String> validNameSet = getValidInputNodeNamesInStandardMode(sources, transforms);
+        for (NodeRelation relation : relations) {
+            List<String> inputs = relation.getInputs();
+            for (int index = 0; index < inputs.size(); ++index) {
+                String inputName = inputs.get(index);
+                if (!(validNameSet.contains(inputName))) {
```
yunqingmoswu, line 196: !(validNameSet.contains(inputName)) -> !validNameSet.contains(inputName)

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
```diff
@@ -137,6 +148,58 @@ private GroupInfo getGroupInfo(InlongGroupInfo groupInfo, List<InlongStreamInfo>
...
+    }
+
+    private void adjustNodeRelations(List<NodeRelation> relations, InlongGroupInfo groupInfo,
+            List<StreamSource> sources, List<TransformResponse> transforms) {
+        if (InlongConstants.LIGHTWEIGHT_MODE.equals(groupInfo.getLightweight())) {
+            return;
+        }
+
+        // set relations' input node to mq node when necessary
+        String mqNodeName = sources.get(0).getSourceName();
```
yunqingmoswu, line 190: Why is it get(0)?

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
```diff
@@ -137,6 +148,58 @@ private GroupInfo getGroupInfo(InlongGroupInfo groupInfo, List<InlongStreamInfo>
...
+    }
+
+    private void preprocessTransformList(InlongGroupInfo groupInfo, List<StreamSource> sources,
+            List<TransformResponse> transforms) {
+        if (InlongConstants.LIGHTWEIGHT_MODE.equals(groupInfo.getLightweight())) {
+            return;
+        }
+
+        // set transform fields' origin node to mq node when necessary
+        String mqNodeName = sources.get(0).getSourceName();
```
yunqingmoswu, line 169: Why is it get(0)?

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
```diff
@@ -137,6 +148,58 @@ private GroupInfo getGroupInfo(InlongGroupInfo groupInfo, List<InlongStreamInfo>
...
+        Set<String> validNameSet = getValidInputNodeNamesInStandardMode(sources, transforms);
+        for (TransformResponse transform : transforms) {
+            for (StreamField field : transform.getFieldList()) {
+                String originNodeName = field.getOriginNodeName();
+                if (!(validNameSet.contains(originNodeName))) {
```
yunqingmoswu, line 174: !(validNameSet.contains(originNodeName)) -> !validNameSet.contains(originNodeName)

yunqingmoswu (commented):
Should the relation between Agent source and MQ source be found instead of MQSources.get(0)

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/resource/sort/DefaultSortConfigOperator.java
woofyzhao: In standard mode there is exactly one mq source for each stream at the moment. 
Its source name is the corresponding stream id.

woofyzhao:
> Should the relation between Agent source and MQ source be found instead of MQSources.get(0)

The source object here, when understand mode, is already the MQ source found. Currently there is exactly one mq source for each stream.

yunqingmoswu (approved):

healchow (approved):

PR #6059, opened 2022-09-29
[INLONG-6058][DataProxy] Add parameter checking and optimize client creation logic

### LINKED ISSUE #6058: [Bug][DataProxy] Add parameter checking and optimize client creation logic
### What happened

In the return result of getConfig, if the configuration item is null, the comparison will fail; at the same time, in the Pulsar client creation structure, the deletion and addition are optimized to avoid the creation failure.

If the value is null, it is considered consistent when compared with the string "null"

### How to reproduce

return null value in getConfig result

### Environment

Linux

### InLong version

master

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #6058

### DIFF
--- a/inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/ConfigManager.java
+++ b/inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/ConfigManager.java
@@ -159,14 +159,9 @@ public class ConfigManager {
         // add new configure records
         for (Map.Entry<String, String> entry : result.entrySet()) {
             String oldValue = tmpHolder.put(entry.getKey(), entry.getValue());
-            if (entry.getValue() == null) {
-                if (!Objects.equals(oldValue, "null")) {
-                    changed = true;
-                }
-            } else {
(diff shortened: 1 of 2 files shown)


baomingyu (approved):
+1

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/ConfigManager.java
```diff
@@ -157,8 +159,14 @@ private boolean updatePropertiesHolder(Map<String, String> result,
...
         for (Map.Entry<String, String> entry : result.entrySet()) {
             String oldValue = tmpHolder.put(entry.getKey(), entry.getValue());
-            if (!ObjectUtils.equals(oldValue, entry.getValue())) {
-                changed = true;
+            if (entry.getValue() == null) {
+                if (!Objects.equals(oldValue, "null")) {
+                    changed = true;
+                }
+            } else {
+                if (!ObjectUtils.equals(oldValue, entry.getValue())) {
```
healchow, line 167:
```suggestion
                if (!Objects.equals(oldValue, entry.getValue())) {
```

healchow (approved):

PR #6072, opened 2022-09-30
[INLONG-6071][Manager] Add doris sink for manager

Yizhou-Yang:
initial doris sink, need connector implementation for further detail


Yizhou-Yang:
- [ ]

PR #6077, opened 2022-09-30
[INLONG-6076][Audit] Bugs in the ClickHouseService.processOutput()

### LINKED ISSUE #6076: [Bug][Audit] Bugs in the ClickHouseService.processOutput()
### What happened

In the ClickHouseService.processOutput() function[a], there are several places that need to be optimized:
1. data needs to read batchQueue data inside the while loop;
2. After exiting the while loop, you need to judge whether there is still unwritten data, and then do the commit operation

a. https://github.com/apache/inlong/blob/master/inlong-audit/audit-store/src/main/java/org/apache/inlong/audit/service/ClickHouseService.java#L106
企业微信截图_16645225199676

fix bugs

### How to reproduce

see code

### Environment

Linux

### InLong version

master

### InLong Component

InLong Audit

### Code of Conduct

gosonzhang:
- Fixes #6076

### DIFF
--- a/inlong-audit/audit-store/src/main/java/org/apache/inlong/audit/service/ClickHouseService.java
+++ b/inlong-audit/audit-store/src/main/java/org/apache/inlong/audit/service/ClickHouseService.java
@@ -99,10 +99,13 @@ public class ClickHouseService implements InsertData, AutoCloseable {
             return;
         }
         // output
-        try (PreparedStatement pstat = this.conn.prepareStatement(INSERT_SQL)) {
-            // insert data
-            ClickHouseDataPo data = this.batchQueue.poll();
+        PreparedStatement pstat = null;
+        try {
             int counter = 0;
+            // get prepare statement
+            pstat = this.conn.prepareStatement(INSERT_SQL);
+            // output data to clickhouse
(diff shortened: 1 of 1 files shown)


inlong-audit/audit-store/src/main/java/org/apache/inlong/audit/service/ClickHouseService.java
```diff
@@ -99,10 +99,13 @@ private void processOutput() {
             return;
         }
         // output
-        try (PreparedStatement pstat = this.conn.prepareStatement(INSERT_SQL)) {
```
healchow, line 102: Suggested using the `try-with-resource`, because the `PreparedStatement` implements the `AutoCloseable` interface, it will be closed automatically, we don't need to write additional `finally` block to close it.

gosonzhang: Ok, you are right

healchow (approved):

doleyzi (approved):

dockerzhang (approved):

PR #6325, opened 2022-10-29
[INLONG-6324][DataProxy] Optimize DataProxy's message processing on the source side

### LINKED ISSUE #6324: [Improve][DataProxy] Optimize DataProxy's message processing on the source side
### Description

When analyzing the implementation of DataProxy recently, I felt that the following places need to be dealt with:
1. FILE_CHECK_DATA and MINUTE_CHECK_DATA type messages are not used, we should reject them, if we really need to add them later;
2. The MSG_BIN_HEARTBEAT type does not return a response message, and this message should not need to be passed to MQ;
3. MSG_UNKNOWN message type should be discarded;
4. When groupId and streamId do not exist, it should be rejected or sent to the default topic according to the configuration to increase the flexibility of message processing
5. Add the error reason for the error message. Due to the problem of the protocol definition, the error can only be added to the attribute. The receiving-side SDK should obtain and process such a clear error message

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #6324

### DIFF
--- a/inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyErrCode.java
+++ b/inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyErrCode.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.common.enums;
+
+public enum DataProxyErrCode {
+
+    ERR_CODE_SUCCESS(0, "Ok"),
+
+    ERR_CODE_UNSUPPORTED_MSGTYPE(1, "Unsupported msgType"),
+    ERR_CODE_EMPTY_MSG(2, "Empty message"),
+    ERR_CODE_UNSUPPORTED_EXTENDFIELD_VALUE(3,
+            "Unsupported extend field value"),
+    ERR_CODE_UNCONFIGURED_GROUPID_OR_STREAMID(4,
+            "Un-configured groupId or streamId"),
+
+    ERR_CODE_UNKNOWN(Integer.MAX_VALUE, "Unknown error");
+
+    private final int errCode;
+    private final String errMsg;
+
+    DataProxyErrCode(int errorCode, String errorMsg) {
+        this.errCode = errorCode;
+        this.errMsg = errorMsg;
+    }
+
+    public static DataProxyErrCode valueOf(int value) {
+        for (DataProxyErrCode msgErrCode : DataProxyErrCode.values()) {
+            if (msgErrCode.getErrCode() == value) {
+                return msgErrCode;
+            }
+        }
+
+        return ERR_CODE_UNKNOWN;
+    }
+
+    public int getErrCode() {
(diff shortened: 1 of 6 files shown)


inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyErrCode.java
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.common.enums;
+
+public enum DataProxyErrCode {
+
+    ERR_CODE_SUCCESS(0, "Ok"),
```
healchow, line 23: Personally, I don't think it is necessary to add the `ERR_CODE` prefix, because the enumeration class name already indicates that all its instances are about ERROR CODE.

gosonzhang: This writing is just the name of the enumeration, and it is only a numerical value when it is interacted, and only a habit problem

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/utils/MessageUtils.java
```diff
@@ -56,33 +63,193 @@ public static boolean isSyncSendForOrder(Event event) {
...
+     * Build heartbeat response message ByteBuf
+     *
+     * @param attrs     the attribute string
+     * @param version   the version
+     * @param loadValue the node load value
+     * @return ByteBuf
+     */
+    private static ByteBuf buildHBRspPackage(String attrs, byte version, int loadValue) {
+        // calculate total length
+        int binTotalLen = 1 + 4 + 1 + 4 + 2 + 2 + 2;
```
EMsnap, line 234: guess it's nessasary to explain the numbers for different part here

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/utils/MessageUtils.java
```diff
@@ -56,33 +63,197 @@ public static boolean isSyncSendForOrder(Event event) {
...
 
     /**
-     * Convert String to ByteBuf
+     *  process response to client
+     * @param commonAttrMap attribute map
+     * @param resultMap     result map
+     * @param remoteChannel client channel
+     * @param msgType       the message type
+     */
+    public static void returnSourceRspPackage(Map<String, String> commonAttrMap,
```
pocozh, line 72: suggest to remove "return" in the method name

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/pulsar/PulsarClientService.java
```diff
@@ -243,7 +242,7 @@ private void sendResponse(OrderEvent orderEvent, String inlongGroupId, String in
                     logger.debug("order message rsp: seqId = {}, inlongGroupId = {}, inlongStreamId = {}", sequenceId,
                             inlongGroupId, inlongStreamId);
                 }
-                ByteBuf binBuffer = MessageUtils.getResponsePackage("", MsgType.MSG_BIN_MULTI_BODY, sequenceId);
+                ByteBuf binBuffer = MessageUtils.buildBinMsgRspPackage("", sequenceId);
```
pocozh, line 245: Suggest to add `dpIp` attribute here.

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/utils/MessageUtils.java
gosonzhang: Ok

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/pulsar/PulsarClientService.java
gosonzhang: DONE

pocozh (approved):

healchow (approved):

EMsnap (approved):

PR #6351, opened 2022-11-01
[INLONG-6329][DataProxy] Add proxy-send mode message logic

### LINKED ISSUE #6329: [Improve][DataProxy] Add proxy-send mode message logic
### Description

At present, there are 2 modes of message processing of DataProxy:
1. The source side returns success or failure after receiving the message, and then sends the message through the channel and sink. Even if the channel and sink processing fails, the processing result will not be returned to the Client;
2. The source side does not return success or failure immediately after receiving the message, but waits until the sink sends it and then returns the success or failure result to the client, but this is done in an orderly synchronous sending mode;

Based on the existing sending mode, we want to build a third type:
3. The source side does not return success or failure immediately after receiving the message, but waits until the sink sends it and then returns the success or failure result to the client. The sink returns the real sending success or failure message to the Client

In order to distinguish the messages required by this type of service, we need the SDK to add "proxySend=true" to the properties of the message to indicate that this type of message requires DataProxy to process the message in the third way.

### InLong Component

InLong DataProxy, InLong SDK

### Code of Conduct

gosonzhang:
- Fixes #6329


inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/consts/AttributeConstants.java
```diff
@@ -88,6 +88,8 @@ public class AttributeConstants {
 
     public static final String MESSAGE_SYNC_SEND = "syncSend";
 
+    public static final String MESSAGE_PROXY_SEND = "proxySend";
```
pocozh, line 91: It's better move these to common module, they are also used by`dataproxy-sdk`

inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyErrCode.java
```diff
@@ -27,7 +27,22 @@ public enum DataProxyErrCode {
     UNSUPPORTED_EXTENDFIELD_VALUE(3,
             "Unsupported extend field value"),
     UNCONFIGURED_GROUPID_OR_STREAMID(4,
-            "Un-configured groupId or streamId"),
+            "Unconfigured groupId or streamId"),
+    PUT_EVENT_TO_CHANNEL_FAILURE(5,
+            "Put event to Channels failure"),
+    TOPIC_IS_BLANK(6,
+            "Topic is null"),
+    NO_AVAILABLE_RPODUCERINFO(7,
```
pocozh, line 35: typo, NO_AVAILABLE_PRODUCERINFO

gosonzhang: done

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/consts/AttributeConstants.java
gosonzhang: ok, you're right!

pocozh (approved):

EMsnap (approved):

inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyErrCode.java
```diff
@@ -27,7 +27,22 @@ public enum DataProxyErrCode {
...
+            "Put event to Channels failure"),
+    TOPIC_IS_BLANK(6,
+            "Topic is null"),
+    NO_AVAILABLE_PRODUCERINFO(7,
+            "No available producer info"),
+    PRODUCER_IS_NULL(8,
+            "Producer is null"),
+    SEND_REQUEST_TO_PULSAR_FAILURE(9,
+            "Send requet to Pulsar failure"),
+    PULSAR_RETURN_ERROR(10,
```
healchow, line 41: I think just returning the `MQ_RETURN_ERROR` is enough, there is no need to specify the actual MQ type.

gosonzhang: ok

healchow (approved):

PR #6630, opened 2022-11-25
[INLONG-6629][GIT] Add .gitattributes to format the files

dockerzhang:
### Prepare a Pull Request
*(Change the title refer to the following example)*

- Title Example: [INLONG-XYZ][Component] Title of the pull request

*(The following *XYZ* should be replaced by the actual [GitHub Issue](https://github.com/apache/inlong/issues) number)*

- Fixes #6629

### Motivation

Declare files that will always have LF line endings on checkout

PR #6634, opened 2022-11-25
Revert "[INLONG-6629][GIT] Add .gitattributes to format the files"

dockerzhang:
Reverts apache/inlong#6630

some files affected may cause many UT fails.

企业微信截图_dd6c5082-0f13-4f87-8fc5-38753d669312


healchow:
Fixed by #6632.

PR #6871, opened 2022-12-13
[INLONG-6860][Manager] Add open APIs related to inlong basic operations

### LINKED ISSUE #6860: [Improve][Manager] Add open APIs related to inlong basic operations
### Description

During use, some users asked whether the system can add related content through API instead of web, such as adding, deleting, modifying and checking related content of inlong-group, inlong-stream, and inlong-sink, can it be done through API?

Analyzing this requirement is necessary for some users, so that it is convenient for the business to embed the process into the system without connecting people. The problem is how to manage between systems and the logic of the current page How to unify, considering that this is a new requirement, temporarily make two parallel logics, and after the relevant requirements are detailed, we will merge them into one API again.

### InLong Component

InLong Manager

### Code of Conduct

gosonzhang:
- Fixes #6860

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
@@ -29,6 +29,7 @@ import org.apache.inlong.manager.common.auth.SecretTokenAuthentication;
 import org.apache.inlong.manager.common.consts.InlongConstants;
 import org.apache.inlong.manager.common.enums.ErrorCodeEnum;
 import org.apache.inlong.manager.common.enums.GroupStatus;
+import org.apache.inlong.manager.common.enums.UserTypeEnum;
 import org.apache.inlong.manager.common.exceptions.BusinessException;
 import org.apache.inlong.manager.common.util.CommonBeanUtils;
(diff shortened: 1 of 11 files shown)


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
```diff
@@ -170,6 +173,29 @@ public String save(InlongGroupRequest request, String operator) {
         return groupId;
     }
 
+    @Override
+    @Transactional(rollbackFor = Throwable.class)
+    public String save(InlongGroupRequest request, UserInfo opInfo) {
```
woofyzhao, line 178: Can we reuse the old method since most of the codes are identical?

gosonzhang: Yes, business logic and page are currently consistent

This idea is to separate the page from the open api first, and implement the call logic separately. After completing the subsequent implementation, we will see if they are the same in the end: if they are consistent, we will merge them into one; if they are inconsistent, we will merge them into one. deal with separately

vernedeng (approved):
lgtm

woofyzhao (approved):

healchow (approved):
LGTM

dockerzhang (approved):

PR #7157, opened 2023-01-05
[INLONG-7156][Agent] Support directly sending raw file data

### LINKED ISSUE #7156: [Improve][Agent] Support directly sending raw file data
### Description

_No response_

### InLong Component

InLong Agent

### Code of Conduct

pocozh:
### Prepare a Pull Request

- [INLONG-7156][Agent] Support directly sending raw file data

- Fixes #7156 

### Motivation

### Modifications

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/file/FileReaderOperator.java
+++ b/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/file/FileReaderOperator.java
@@ -21,11 +21,11 @@ import com.google.gson.Gson;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.inlong.agent.conf.JobProfile;
+import org.apache.inlong.agent.except.FileException;
 import org.apache.inlong.agent.message.DefaultMessage;
 import org.apache.inlong.agent.metrics.audit.AuditUtils;
 import org.apache.inlong.agent.plugin.Message;
 import org.apache.inlong.agent.plugin.Validator;
-import org.apache.inlong.agent.except.FileException;
 import org.apache.inlong.agent.plugin.sources.reader.AbstractReader;
 import org.apache.inlong.agent.plugin.utils.FileDataUtils;
 import org.apache.inlong.agent.plugin.validator.PatternValidator;
@@ -104,6 +104,7 @@ public class FileReaderOperator extends AbstractReader {
 
     private final BlockingQueue<String> queue = new LinkedBlockingQueue<>(CACHE_QUEUE_SIZE);
     private final StringBuffer sb = new StringBuffer();
+    private boolean needMetaData = false;
 
     public FileReaderOperator(File file, int position) {
         this(file, position, "");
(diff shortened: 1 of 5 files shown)


inlong-agent/agent-plugins/src/test/java/org/apache/inlong/agent/plugin/task/TestTextFileTask.java
```diff
@@ -180,7 +182,7 @@ public void testReadFull() throws IOException {
         jobProfile.set(JobConstants.JOB_DIR_FILTER_PATTERNS, file.getAbsolutePath());
         jobProfile.set(JobConstants.JOB_TASK_BEGIN_WAIT_SECONDS, String.valueOf(0));
         jobProfile.set(JobConstants.JOB_FILE_CONTENT_COLLECT_TYPE, DataCollectType.FULL);
-
+        jobProfile.set(JOB_FILE_META_ENV_LIST, KUBERNETES);
```
thesumery, line 185: Pls add an unit test for non metadata text read.

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/file/FileReaderOperator.java
```diff
@@ -278,6 +282,9 @@ public void resiterMeta(JobProfile jobConf) {
             return;
         }
         String[] env = jobConf.get(JOB_FILE_META_ENV_LIST).split(COMMA);
+        if (env.length > 0) {
```
healchow, line 285: Why is metadata required when there are more than one environment?

pocozh: > Why is metadata required when there are more than one environment?

MetaData is used for some special scenes, such as k8s log.

healchow: If the value of `JOB_FILE_META_ENV_LIST` is "local" or others, but not "k8s", then `needMetaData` will also be true. Is this as expected?

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/file/FileReaderOperator.java
```diff
@@ -104,6 +104,7 @@ public class FileReaderOperator extends AbstractReader {
 
     private final BlockingQueue<String> queue = new LinkedBlockingQueue<>(CACHE_QUEUE_SIZE);
     private final StringBuffer sb = new StringBuffer();
+    private boolean needMetaData = false;
```
healchow, line 107: `Metadata` is a complete word, so don't need to write it as `MetaData`.

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/file/FileReaderOperator.java
pocozh: > If the value of `JOB_FILE_META_ENV_LIST` is "local" or others, but not "k8s", then `needMetaData` will also be true. Is this as expected?

Okay, you are right

dockerzhang (approved):

healchow (approved):
LGTM

PR #7170, opened 2023-01-05
[INLONG-7169][Manager] Optimize OpenDataNodeController implementation

### LINKED ISSUE #7169: [Improve][Manager] Optimize OpenDataNodeController implementation
### Description

The parent issue is :https://github.com/apache/inlong/issues/7168

Based on the problems mentioned by the parent class requirements, optimize the implementation of OpenDataNodeController

### InLong Component

InLong Manager

### Code of Conduct

gosonzhang:
- Fixes #7169

### DIFF
--- a/inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
+++ b/inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
@@ -85,8 +85,8 @@ public class ElasticsearchDataNodeDTO {
         try {
             return JsonUtils.parseObject(extParams, ElasticsearchDataNodeDTO.class);
         } catch (Exception e) {
-            LOGGER.error("Failed to extract additional parameters for Elasticsearch data node: ", e);
-            throw new BusinessException(ErrorCodeEnum.GROUP_INFO_INCORRECT.getMessage());
+            throw new BusinessException(ErrorCodeEnum.GROUP_INFO_INCORRECT,
+                    String.format("Failed to parse extParams for Elasticsearch node: %s", e.getMessage()));
         }
     }
 
--- a/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenDataNodeController.java
+++ b/inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenDataNodeController.java
@@ -19,7 +19,7 @@ package org.apache.inlong.manager.web.controller.openapi;
 
 import org.apache.inlong.manager.common.enums.OperationType;
 import org.apache.inlong.manager.common.validation.SaveValidation;
-import org.apache.inlong.manager.common.validation.UpdateValidation;
+import org.apache.inlong.manager.common.validation.UpdateByIdValidation;
 import org.apache.inlong.manager.pojo.common.Response;
 import org.apache.inlong.manager.pojo.node.DataNodeInfo;
 import org.apache.inlong.manager.pojo.node.DataNodePageRequest;
@@ -77,7 +77,7 @@ public class OpenDataNodeController {
     @PostMapping(value = "/node/update")
     @ApiOperation(value = "Update data node")
     @OperationLog(operation = OperationType.UPDATE)
-    public Response<Boolean> update(@Validated(UpdateValidation.class) @RequestBody DataNodeRequest request) {
+    public Response<Boolean> update(@Validated(UpdateByIdValidation.class) @RequestBody DataNodeRequest request) {
         return Response.success(dataNodeService.update(request, LoginUserUtils.getLoginUser()));
     }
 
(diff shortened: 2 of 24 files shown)


inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
```diff
@@ -85,8 +85,8 @@ public static ElasticsearchDataNodeDTO getFromJson(@NotNull String extParams) {
         try {
             return JsonUtils.parseObject(extParams, ElasticsearchDataNodeDTO.class);
         } catch (Exception e) {
-            LOGGER.error("Failed to extract additional parameters for Elasticsearch data node: ", e);
```
healchow, line 88: Does the `e.getMessage()` show the detail of the parse error? Such as which field was parsed failed.

If not, then I think this error log is necessary.

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenDataNodeController.java
```diff
@@ -71,24 +70,21 @@ public Response<List<DataNodeInfo>> list(@RequestBody DataNodePageRequest reques
...
-    @RequiresRoles(value = UserRoleCode.ADMIN)
-    public Response<Integer> save(@Validated @RequestBody DataNodeRequest request) {
+    public Response<Integer> save(@Validated(SaveValidation.class) @RequestBody DataNodeRequest request) {
         return Response.success(dataNodeService.save(request, LoginUserUtils.getLoginUser()));
     }
 
     @PostMapping(value = "/node/update")
-    @OperationLog(operation = OperationType.UPDATE)
     @ApiOperation(value = "Update data node")
-    @RequiresRoles(value = UserRoleCode.ADMIN)
```
healchow, line 82: Excuse me, but why remove the roles check?

fuweng11 (approved):

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenDataNodeController.java
gosonzhang: The role check is performed in the open api, and this will be strengthened in the future to make it more detailed

inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
gosonzhang: API calls will be more frequent. If every request is logged, the manager node will explode due to too much log output
Then the exception of this block is the json encoding and decoding operation of the extParams content. If it fails, the caller will provide the specific content.

woofyzhao (approved):

inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
healchow: I did not understand that "If it fails, the caller will provide the specific content."

This extParams parameter is saved in the DB. In case of wrong saving in the DB, or the target conversion class changes, which leads to parsing failure, can you get the specific information of extParams from "e.getMessage"?

If you can't get it, you need to check the specific information in the DB according to the previous series of logs. This is undoubtedly more difficult.

gosonzhang: The modification of this place is to allow the caller to analyze the problem, rather than the system administrator to go to the log to analyze the problem

Specifically for this error, the error is the parameter decoding error of extParams, mainly two types, one is that the data passed by the caller is not in json format, and the other is the content of the class that does not match the class specified by the parser.

As the caller, it is clear what content is passed. Based on this basis, the error code and error information provided by the assistant are very clear about which field has an error. Through the specified parsing error information and the original data of the call, the call The operator can analyze the problem by himself, without the need for the system maintainer to view the log, that is, this log output is unnecessary.

dockerzhang (approved):

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenDataNodeController.java
healchow: Okay.

inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/node/es/ElasticsearchDataNodeDTO.java
healchow: I think our discussion may be a little inconsistent.
**It is correct to throw an exception, but my question is, can "e.getMessage" throw the details of the error to the caller?**

gosonzhang: You can simulate and throw an exception directly at the above modification point, to see if the error message of the exception can be obtained in the response

PR #7230, opened 2023-01-13
[INLONG-7229][Manager] Add checks for unmodifiable field values

### LINKED ISSUE #7229: [Improve][Manager] Add checks for unmodifiable field values
### Description

The parent issue is :https://github.com/apache/inlong/issues/7168

Based on the changes in https://github.com/apache/inlong/issues/7220, I found that during the update operation, there is no effective verification for the value check of unmodifiable fields. For example, sinkType in OpenStreamSinkController is an unmodifiable field, the original implementation Direct use without checking will cause implementation bugs; another example is the version field, which should be checked when it receives a request to reduce modification conflicts

This PR supplements the modified api in the parent issue, and the unmodified api will be improved in the subsequent sub-issue.

### InLong Component

InLong Manager

### Code of Conduct

gosonzhang:
- Fixes #7229

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
@@ -68,7 +68,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.stream.Collectors;
 
 /**
@@ -152,10 +151,6 @@ public class StreamSinkServiceImpl implements StreamSinkService {
     public Integer save(SinkRequest request, UserInfo opInfo) {
         // check request parameter
         checkSinkRequestParams(request);
-        // check operator info
-        if (opInfo == null) {
-            throw new BusinessException(ErrorCodeEnum.LOGIN_USER_EMPTY);
-        }
         InlongGroupEntity entity = groupMapper.selectByGroupId(request.getInlongGroupId());
         if (entity == null) {
             throw new BusinessException(ErrorCodeEnum.GROUP_NOT_FOUND,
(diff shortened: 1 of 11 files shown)


fuweng11 (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
```diff
@@ -372,24 +372,53 @@ public List<? extends StreamSink> listByCondition(SinkPageRequest request, UserI
...
-        this.checkParams(request);
-        Preconditions.checkNotNull(request.getId(), ErrorCodeEnum.ID_IS_EMPTY.getMessage());
-
-        // Check if it can be modified
-        String groupId = request.getInlongGroupId();
-        String streamId = request.getInlongStreamId();
-        String sinkName = request.getSinkName();
-        groupCheckService.checkGroupStatus(groupId, operator);
-
+        if (request == null) {
```
healchow, line 375: For the "Service" layer method, it is recommended to put the parameter verification in the "Controller", because the external access is only the "Controller".

The calls of other "Service" layers in the system are all trusted data, and no more judgment is needed.

healchow (changes requested):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
```diff
@@ -372,24 +372,53 @@ public List<? extends StreamSink> listByCondition(SinkPageRequest request, UserI
...
+        }
+        if (request.getId() == null) {
+            throw new BusinessException(ErrorCodeEnum.ID_IS_EMPTY);
+        }
+        StreamSinkEntity curEntity = sinkMapper.selectByPrimaryKey(request.getId());
+        if (curEntity == null) {
+            throw new BusinessException(ErrorCodeEnum.SINK_INFO_NOT_FOUND);
+        }
+        // check record version
+        if (!Objects.equals(curEntity.getVersion(), request.getVersion())) {
```
healchow, line 387: Finally, for the verification of fields that are not allowed to be modified, please extract them into a special method for verifying parameters, which is more object-oriented and also prevents too much code in a single method.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
gosonzhang: Agree

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/sink/StreamSinkServiceImpl.java
gosonzhang: Agree

healchow (approved):
LGTM

dockerzhang (approved):

PR #7255, opened 2023-01-17
[INLONG-7254][Manager] Fix config error when InlongGroupId is in the process of switching

### LINKED ISSUE #7254: [Bug][Manager] Fix config error when InlongGroupId is in the process of switching
### What happened

When a InlongGroupId is in the process of switching, the config of source cluster and target cluster will be mixed up.
Which doubles the data consuming from MQ.

Only consume the clusters under the specify tag, instead of both target tag and backup tag.

### How to reproduce

Switch one InlongGroupId

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

vernedeng:
### Prepare a Pull Request

- Fixes #7254 

### Motivation

When a InlongGroupId is in the process of switching, the config of source cluster and target cluster will be mixed up.
Which doubles the data consuming from MQ.

The solution is to check if the target sort cluster is tagged. If so, fiter the MQs under this tag; otherwise, get all MQs of both source and target MQs.

### Modifications

### Verifying this change

- [ ] This change added tests and can be verified as follows: Swiching in tagged/untagged cluster.

### Documentation

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/SortSourceServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/SortSourceServiceImpl.java
@@ -17,6 +17,7 @@
 
 package org.apache.inlong.manager.service.core.impl;
 
+import com.google.common.base.Preconditions;
 import com.google.gson.Gson;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.apache.commons.lang3.StringUtils;
@@ -44,6 +45,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 import javax.annotation.PostConstruct;
+import javax.swing.text.html.Option;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -51,6 +53,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


woofyzhao (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/SortSourceServiceImpl.java
```diff
@@ -285,8 +290,12 @@ private void parseAll() {
     }
 
     private Map<String, CacheZone> parseCacheZones(
+            String clusterName,
             List<SortSourceStreamSinkInfo> sinkList) {
 
+        Preconditions.checkNotNull(sortClusters.get(clusterName));
```
healchow, line 296: Suggest use the `Preconditions` in manager-common module.

vernedeng: Fixed, thx

healchow (approved):
LGTM

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/SortSourceServiceImpl.java
```diff
@@ -296,14 +305,26 @@ private Map<String, CacheZone> parseCacheZones(
 
         // group them by cluster tag.
         Map<String, List<SortSourceStreamSinkInfo>> tag2SinkInfos = sinkInfoList.stream()
+                .filter(sink -> {
+                    if (StringUtils.isBlank(sortClusterTag)) {
+                        return true;
+                    }
+                    return groupInfos.get(sink.getGroupId()).getClusterTag().equals(sortClusterTag);
```
fuweng11, line 312: Maybe there will be NPE?

gosonzhang (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/SortSourceServiceImpl.java
vernedeng: The NPE will be thrown when there is a stream sink of a groupId  but this groupId does not exist in inlong_group table. I will add one more filter condition to avoid this case.

fuweng11 (approved):

dockerzhang (approved):

PR #7279, opened 2023-01-29
[INLONG-7278][Manager] Optimize OpenInLongClusterController implementation

### LINKED ISSUE #7278: [Improve][Manager] Optimize OpenInLongClusterController implementation
### Description

The parent issue is :#7168

Based on the problems mentioned by the parent class requirements, optimize the implementation of OpenInLongClusterController

### InLong Component

InLong Manager

### Code of Conduct

gosonzhang:
- Fixes #7278

Modifications:
1. Adjust the exception throwing content and increase the error code;
2. Adjust the log printing level, from info to debug
3. Adjust the check of whether the field is empty when the ClusterTagRequest, ClusterRequest, and ClusterNodeRequest objects are modified;
4. Fix the bug when inspecting the code

### DIFF
--- a/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterTagEntityMapper.xml
+++ b/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterTagEntityMapper.xml
@@ -84,7 +84,7 @@
                 cluster_tag = #{clusterTag,jdbcType=VARCHAR},
             </if>
             <if test="extParams != null">
-                ext_params = #{extParams,jdbcType=VARCHAR},
+                ext_params = #{extParams,jdbcType=LONGVARCHAR},
             </if>
             <if test="description != null">
                 description = #{description,jdbcType=VARCHAR},
--- a/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterTagEntityMapper.xml
+++ b/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterTagEntityMapper.xml
@@ -77,16 +77,32 @@
         order by modify_time desc
     </select>
 
-    <update id="updateById" parameterType="org.apache.inlong.manager.dao.entity.InlongClusterTagEntity">
+    <update id="updateByIdSelective" parameterType="org.apache.inlong.manager.dao.entity.InlongClusterTagEntity">
         update inlong_cluster_tag
-        set cluster_tag = #{clusterTag,jdbcType=VARCHAR},
-            ext_params  = #{extParams,jdbcType=LONGVARCHAR},
-            description = #{description,jdbcType=VARCHAR},
-            in_charges  = #{inCharges,jdbcType=VARCHAR},
-            status      = #{status,jdbcType=INTEGER},
-            is_deleted  = #{isDeleted,jdbcType=INTEGER},
-            modifier    = #{modifier,jdbcType=VARCHAR},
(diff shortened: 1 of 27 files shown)


inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenInLongClusterController.java
```diff
@@ -68,93 +70,115 @@ public class OpenInLongClusterController {
...
     public Response<ClusterTagResponse> getTag(@PathVariable Integer id) {
+        Preconditions.checkNull(id, ErrorCodeEnum.INVALID_PARAMETER, "tag id cannot be null");
+        Preconditions.checkNull(LoginUserUtils.getLoginUser(), ErrorCodeEnum.LOGIN_USER_EMPTY);
         return Response.success(clusterService.getTag(id, LoginUserUtils.getLoginUser()));
     }
 
     @PostMapping(value = "/cluster/tag/list")
     @ApiOperation(value = "List cluster tags")
     public Response<List<ClusterTagResponse>> listTag(@RequestBody ClusterTagPageRequest request) {
+        Preconditions.checkNull(request, ErrorCodeEnum.INVALID_PARAMETER, "request cannot be null");
```
vernedeng, line 81: I think it should be Preconditions.checkNotNull

vernedeng (changes requested):

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/openapi/OpenInLongClusterController.java
gosonzhang: This is not a problem, the implementation of this function is to check whether the target is null, and throw an exception when the target is null.

There is no problem with the implementation, but the API naming needs to be modified:
Check only expresses the action, not the expected result, so there is ambiguity. A issue will be mentioned later to modify the naming problem

gosonzhang: It will be traced by https://github.com/apache/inlong/issues/7280

healchow (approved):
LGTM

inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterTagEntityMapper.xml
```diff
@@ -77,16 +77,32 @@
...
-            in_charges  = #{inCharges,jdbcType=VARCHAR},
-            status      = #{status,jdbcType=INTEGER},
-            is_deleted  = #{isDeleted,jdbcType=INTEGER},
-            modifier    = #{modifier,jdbcType=VARCHAR},
+        <set>
+            <if test="clusterTag != null">
+                cluster_tag = #{clusterTag,jdbcType=VARCHAR},
+            </if>
+            <if test="extParams != null">
+                ext_params = #{extParams,jdbcType=VARCHAR},
```
fuweng11, line 87: jdbcType=LONGVARCHAR

gosonzhang: done

vernedeng (approved):
LGTM

Issue #7286, opened 2023-01-30
[Bug][Sort] Fix issue of tableidentifier being null when addRow


Yizhou-Yang:
### What happened

java.lang.ArrayIndexOutOfBoundsException: 1
	at org.apache.inlong.sort.doris.table.DorisDynamicSchemaOutputFormat.load(DorisDynamicSchemaOutputFormat.java:742) ~[pool-1-thread-5-1675066174508-test-sort-connector-doris-1.5.0-SNAPSHOT.jar:1.6.0-SNAPSHOT]
	at org.apache.inlong.sort.doris.table.DorisDynamicSchemaOutputFormat.flushSingleTable(DorisDynamicSchemaOutputFormat.java:669) ~[pool-1-thread-5-1675066174508-test-sort-connector-doris-1.5.0-SNAPSHOT.jar:1.6.0-SNAPSHOT]
	at org.apache.inlong.sort.doris.table.DorisDynamicSchemaOutputFormat.flush(DorisDynamicSchemaOutputFormat.java:646) ~[pool-1-thread-5-1675066174508-test-sort-connector-doris-1.5.0-SNAPSHOT.jar:1.6.0-SNAPSHOT]
	at org.apache.inlong.sort.doris.table.DorisDynamicSchemaOutputFormat.lambda$open$1(DorisDynamicSchemaOutputFormat.java:298) ~[pool-1-thread-5-1675066174508-test-sort-connector-doris-1.5.0-SNAPSHOT.jar:1.6.0-SNAPSHOT]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) [?:1.8.0_332]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) [?:1.8.0_332]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) [?:1.8.0_332]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294) [?:1.8.0_332]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_332]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_332]
[...]

PR #7287, opened 2023-01-30
[INLONG-7286][Sort] Fix issue of tableidentifier being null when addRow

Yizhou-Yang:
https://github.com/apache/inlong/issues/7286

------------------------------------------------
Motivation:
An error stack which is arrayIndexoutofbounds. 
After investigation, it is because the dirtyidentifier is null for multiple sink.
This was okay since we dont need the identifier to process data in multiple sink scenarios, but this is not ok for dirty data log output

Modifications:
generated a precise dirtyidentifier in addrow, so that load() call will have the actual identifier instead of null.

### DIFF
--- a/inlong-sort/sort-connectors/doris/src/main/java/org/apache/inlong/sort/doris/table/DorisDynamicSchemaOutputFormat.java
+++ b/inlong-sort/sort-connectors/doris/src/main/java/org/apache/inlong/sort/doris/table/DorisDynamicSchemaOutputFormat.java
@@ -441,7 +441,10 @@ public class DorisDynamicSchemaOutputFormat<T> extends RichOutputFormat<T> {
 
     @SuppressWarnings({"unchecked"})
     private void addRow(RowKind rowKind, JsonNode rootNode, JsonNode physicalNode, JsonNode updateBeforeNode,
-            Map<String, String> physicalData, Map<String, String> updateBeforeData) {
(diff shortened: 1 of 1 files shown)


inlong-sort/sort-connectors/doris/src/main/java/org/apache/inlong/sort/doris/table/DorisDynamicSchemaOutputFormat.java
```diff
@@ -441,7 +441,10 @@ private void addBatch(T row) {
 
     @SuppressWarnings({"unchecked"})
     private void addRow(RowKind rowKind, JsonNode rootNode, JsonNode physicalNode, JsonNode updateBeforeNode,
-            Map<String, String> physicalData, Map<String, String> updateBeforeData) {
+            Map<String, String> physicalData, Map<String, String> updateBeforeData) throws IOException {
+        String tableIdentifier = StringUtils.join(
```
yunqingmoswu, line 445: make sure this can resolve it ?

Yizhou-Yang: it worked in the testing environment. will this introduce new problems?

gong:
PR description should be format: Motivation, Modifications.

gong (approved):

EMsnap (approved):

yunqingmoswu (approved):

PR #7301, opened 2023-02-01
[INLONG-7300][Manager] Replace getRoles() with getAccountType()

### LINKED ISSUE #7300: [Improve][Manager] Replace getRoles() with getAccountType()
### Description

The parent issue: https://github.com/apache/inlong/issues/7168

In the user authorization check of the Open APIs, the getRoles() function of the UserInfo class is used, but the getAccountType() function is used in the current Manager implementation. In order to maintain the consistency of the authentication and authorization logic, the getRoles() calls are replaced with getAccountType()

### InLong Component

InLong Manager

### Code of Conduct

gosonzhang:
Replace getRoles() with getAccountType() in Open APIs

- Fixes #7300

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterServiceImpl.java
@@ -146,7 +146,7 @@ public class InlongClusterServiceImpl implements InlongClusterService {
     @Override
     public Integer saveTag(ClusterTagRequest request, UserInfo opInfo) {
         // only the person in charges can query
-        if (!opInfo.getRoles().contains(UserTypeEnum.ADMIN.name())) {
+        if (!opInfo.getAccountType().equals(UserTypeEnum.ADMIN.getCode())) {
             throw new BusinessException(ErrorCodeEnum.PERMISSION_REQUIRED,
                     "Current user does not have permission to add cluster tag");
         }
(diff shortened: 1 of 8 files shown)


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterServiceImpl.java
```diff
@@ -146,7 +146,7 @@ public Integer saveTag(ClusterTagRequest request, String operator) {
     @Override
     public Integer saveTag(ClusterTagRequest request, UserInfo opInfo) {
         // only the person in charges can query
-        if (!opInfo.getRoles().contains(UserTypeEnum.ADMIN.name())) {
+        if (!opInfo.getAccountType().equals(UserTypeEnum.ADMIN.getCode())) {
```
healchow, line 149: Semantically, account type has less meaning than user role. It is more common to manage permissions based on user roles.
So I think this change is not a good practice.

gosonzhang: I agree with what you said, but the current practice has not yet completed authentication and authorization based on roles, so first judge according to the current implementation based on AccountType, and then deal with it according to roles after subsequent improvements, or it will become a functional bug

gosonzhang: This PR only adjusts the newly added implementation to be consistent with the current system practice, and authorization issues according to roles will be handle later @healchow

fuweng11 (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterServiceImpl.java
healchow: Okay, looking forward to your next PR 😄

healchow (approved):
LGTM

PR #7312, opened 2023-02-02
[INLONG-7311][Sort] Fix Doris StreamLoad unable to archive dirtydata

### LINKED ISSUE #7311: [Bug] [Sort] Doris StreamLoad unable to archive dirtydata
### What happened

Two bugs for doris single table and multiple sink, I think that they share a common cause.

image

image

fix both bugs and archive dirty data properly

### How to reproduce

do a doris single/multiple data archive with exception.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Sort

### Code of Conduct

Yizhou-Yang:
### Prepare a Pull Request

- [INLONG-7311][Sort] Fix Doris StreamLoad unable to archive dirtydata

- Fixes #7311 

### Motivation
When testing dirty data to doris sink, I found that none of the dirty data is archived, and the metric is always 0 for successfully written and 0 for dirty.

image

I found the cause is that when flush(), the data is written by DorisStreamLoad.load(), and within this method exception will occur.

However, when load() fails, its return value, respContent, will be null since an exception is thrown, and at this time, the current logic will not archive this as dirty data.

### Modifications
Changed the archive strategy so that  when respContent is null (exception during streamload), the unwritten data will be archived as dirtydata if dirtysink is not null.


dockerzhang (approved):

healchow (approved):
LGTM

EMsnap (approved):

PR #7393, opened 2023-02-20
[INLONG-7392][Sort] Refactor Doris single table to solve performance issues 

### LINKED ISSUE #7392: [Improve][Sort] Refactor Doris single table to solve performance issues
### Description

Currently, Doris single table has a shockingly low throughput (1700 records/s, where it should be 50000). I need to ensure that its performance is good enough.

### InLong Component

InLong Sort

### Code of Conduct

Yizhou-Yang:
- Fixes #7392

### Motivation

Currently, Doris single table has low throughput (1700 records/s, where it should be 50000). I need to ensure that its performance is good enough.

The cause of the issue is because Doris multiple logic and single table logic are not actually compatible. Instead of doing a huge modification on the base class, I choose to make them into two separate classes, so the logic is simpler in each of the two classes.

### Modifications
To fix this issue, I am isolating single and multiple logic for Doris, and refactor the internal logic for doris single table. DirtyHelper, Inlong metric and etc. are still being kept.

1) create DorisSingleOutputFormat to handle single table logic without repeated access to the single identifier in batchmap
2) create relevant interfaces and builders based on multipleSink, so that DorisSingleOutputFormat is used instead of DorisDynamicSchemaOutputFormat when handling single table logics
3) remove single table logic processing from DorisDynamicSchemaOutputFormat*

### DIFF
--- a/inlong-sort/sort-connectors/doris/src/main/java/org/apache/inlong/sort/doris/internal/DorisOutputFormat.java
+++ b/inlong-sort/sort-connectors/doris/src/main/java/org/apache/inlong/sort/doris/internal/DorisOutputFormat.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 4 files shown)


Yizhou-Yang:
comment: this fix is unnecessary. could have been a one-line fixer.

Issue #7419, opened 2023-02-23
[Bug][Agent] The connector of MQTT is unable to connect


bluewang:
### What happened

![image](https://user-images.githubusercontent.com/88174078/220863431-af7cbd41-3114-4f13-9056-625a2df75961.png)

![image](https://user-images.githubusercontent.com/88174078/220863477-3af128e2-ea84-4eee-8ceb-671a091a4472.png)

### What you expected to happen

Fix the agent of the MQTT connector

### How to reproduce

user the MQTT source

### InLong version

master

### InLong Component

InLong Agent

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7420, opened 2023-02-23
[INLONG-7419][Agent] Fix error of MQTT connector

bluewang:
### Prepare a Pull Request

- Fixes https://github.com/apache/inlong/issues/7419

### Verifying this change
image

### DIFF
--- a/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/MqttReader.java
+++ b/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/MqttReader.java
@@ -119,8 +119,8 @@ public class MqttReader extends AbstractReader {
 
                     @Override
                     public void connectionLost(Throwable cause) {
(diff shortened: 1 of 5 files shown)


dockerzhang (approved):

haifxu (approved):

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/reader/MqttReader.java
```diff
@@ -116,8 +119,8 @@ private void connect() {
 
                     @Override
                     public void connectionLost(Throwable cause) {
-                        LOGGER.info("the mqtt connection is lost, try to reconnect. jobId:{},serverURI:{},clientId:{}",
-                                instanceId, serverURI, clientId);
+                        LOGGER.error("the mqtt jobId:{}, serverURI:{}, connection lost, {} ", instanceId,
+                                serverURI, cause.getMessage());
```
healchow, line 123: Do we need to print all stacks of the exception? Maybe using `cause` instead of `cause.getMessage()` is better if required.

healchow (approved):

Issue #7455, opened 2023-02-28
[Bug][Sort] Fix dirty data archival format issues in Iceberg connector


lordcheng10:
### What happened

Start a MySQL synchronization task with Iceberg and insert a type mismatch of dirty data, then observe the format of the dirty data content after archiving.

### What you expected to happen

Start a mysql synchronization task for Iceberg and insert a type mismatch of dirty data, the archived dirty data should contain data content，eg: [time] [database] [table] [data]

### How to reproduce

Start a MySQL synchronization task with Iceberg and insert a type mismatch of dirty data, then observe the format of the dirty data content after archiving.

### InLong version

master

### InLong Component

InLong Sort

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #7456, opened 2023-02-28
[INLONG-7455][Sort] Fix dirty data archival format issues in Iceberg connector

lordcheng10:
### Prepare a Pull Request
- Fixes https://github.com/apache/inlong/issues/7455

### Motivation
Start a mysql synchronization task for Iceberg and insert a type mismatch of dirty data, the archived dirty data should contain data content，eg: [time] [database] [table] [data]

But the test found that it does not contain data：

企业微信截图_4ba5c70c-09d2-4093-8bf9-1e1cf8791c6f

### Modifications
When the types do not match, the dirty data type passed to the handleDirtyData method should not be in json format, but should be converted into a RowData object:
https://github.com/apache/inlong/blob/08887b1dc240c015775cfc3c16b5ec8666833a78/inlong-sort/sort-connectors/iceberg/src/main/java/org/apache/inlong/sort/iceberg/sink/multiple/DynamicSchemaHandleOperator.java#L314-L316


inlong-sort/sort-connectors/iceberg/src/main/java/org/apache/inlong/sort/iceberg/sink/multiple/DynamicSchemaHandleOperator.java
```diff
@@ -314,7 +314,11 @@ private void handleSchemaInfoEvent(TableIdentifier tableId, Schema schema) {
                                 LOG.warn("Ignore table {} schema change, old: {} new: {}.",
                                         tableId, dataSchema, latestSchema, e);
                                 blacklist.add(tableId);
-                                handleDirtyData(jsonNode, jsonNode, DirtyType.EXTRACT_ROWDATA_ERROR, e, tableId);
+                                List<RowData> rowDataForDataSchemaList = dynamicSchemaFormat.extractRowData(jsonNode,
```
Yizhou-Yang, line 317: is there a possibility that this line will throw an exception? If so, is it handled properly?

lordcheng10: 1. dynamicSchemaFormat.extractRowData() usually throws an exception when the schema does not match.
2. Use dataSchema for conversion, usually no exception will be thrown;
3. We can also use try catch to output the error log when an exception is thrown? @Yizhou-Yang

Yizhou-Yang (approved):

dockerzhang (approved):

EMsnap (approved):

PR #7578, opened 2023-03-13
[INLONG-7577][Manager] Fix high CPU usage when the number of StreamSource is too large

### LINKED ISSUE #7577: [Bug][Manager] The number of StreamSource is too large, resulting in high CPU usage
### What happened

When too many agents are deployed, a large number of StreamSource instances are generated when the agent pulls tasks from the manager, leading to a large number of GC.The result is CPU100%.

The manager does not generate too many streamSource instances when pulling tasks.

企业微信截图_d978311a-ebb5-4773-a814-ce0bff687feb

### How to reproduce

More than 100 agents are deployed.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

fuweng11:
### Prepare a Pull Request

- Fixes #7577 

### Motivation

Fix the manager generate too many streamSource instances when pulling tasks.
### Modifications
The previous query was to pull out all Streamsources in the templateId and filter them.
If there were 200 agents, this method would generate 500 + streamSource instances each time, leading to a large amount of gc.
Filtering is now done directly in the mysql layer to avoid generating too many instances and causing gc.

### DIFF
--- a/inlong-manager/manager-dao/src/main/java/org/apache/inlong/manager/dao/mapper/StreamSourceEntityMapper.java
+++ b/inlong-manager/manager-dao/src/main/java/org/apache/inlong/manager/dao/mapper/StreamSourceEntityMapper.java
@@ -41,7 +41,14 @@ public interface StreamSourceEntityMapper {
      */
     StreamSourceEntity selectForAgentTask(Integer id);
 
-    StreamSourceEntity selectExistsByTemplateIdAndIp(@Param("templateId") Integer templateId,
+    /**
+     * Select one sub source by template id and agent ip.
+     *
+     * @param templateId template id
+     * @param agentIp agent ip
+     * @return stream source info
+     */
+    StreamSourceEntity selectOneByTemplatedIdAndAgentIp(@Param("templateId") Integer templateId,
             @Param("agentIp") String agentIp);
 
     /**
--- a/inlong-manager/manager-dao/src/main/java/org/apache/inlong/manager/dao/mapper/StreamSourceEntityMapper.java
+++ b/inlong-manager/manager-dao/src/main/java/org/apache/inlong/manager/dao/mapper/StreamSourceEntityMapper.java
@@ -41,6 +41,9 @@ public interface StreamSourceEntityMapper {
      */
     StreamSourceEntity selectForAgentTask(Integer id);
 
+    StreamSourceEntity selectExistsByTemplateIdAndIp(@Param("templateId") Integer templateId,
+            @Param("agentIp") String agentIp);
+
     /**
      * Query un-deleted sources by the given agentIp.
      */
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
@@ -335,7 +335,7 @@ public class AgentServiceImpl implements AgentService {
         List<StreamSourceEntity> sourceEntities = sourceMapper.selectTemplateSourceByCluster(needCopiedStatusList,
                 Lists.newArrayList(SourceType.FILE), agentClusterName);
         sourceEntities.forEach(sourceEntity -> {
-            StreamSourceEntity subSource = sourceMapper.selectExistsByTemplateIdAndIp(sourceEntity.getId(),
+            StreamSourceEntity subSource = sourceMapper.selectOneByTemplatedIdAndAgentIp(sourceEntity.getId(),
                     agentIp);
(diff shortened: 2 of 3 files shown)


inlong-manager/manager-dao/src/main/java/org/apache/inlong/manager/dao/mapper/StreamSourceEntityMapper.java
```diff
@@ -41,6 +41,9 @@ public interface StreamSourceEntityMapper {
      */
     StreamSourceEntity selectForAgentTask(Integer id);
 
+    StreamSourceEntity selectExistsByTemplateIdAndIp(@Param("templateId") Integer templateId,
```
healchow, line 44: 1. Please add Javadoc for the method.
2. Maybe `selectOneByTemplatedIdAndAgentIp` is more readable.

fuweng11: done.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
```diff
@@ -338,31 +334,25 @@ private void preProcessTemplateFileTask(TaskRequest taskRequest) {
...
-                    fileEntity.setTemplateId(sourceEntity.getId());
-                    fileEntity.setAgentIp(agentIp);
-                    fileEntity.setStatus(SourceStatus.TO_BE_ISSUED_ADD.getCode());
-                    // create new sub source task
-                    sourceMapper.insert(fileEntity);
-                    LOGGER.info("Transform new template task({}) for agent({}) in cluster({}).",
-                            fileEntity.getId(), taskRequest.getAgentIp(), taskRequest.getClusterName());
-                });
+        sourceEntities.forEach(sourceEntity -> {
+            StreamSourceEntity subSource = sourceMapper.selectOneByTemplatedIdAndAgentIp(sourceEntity.getId(),
```
healchow, line 338: Are you sure it's the code here that is causing the high CPU usage?

Only one query is made here, and the amount of data is too large, which should lead to excessive memory usage, and then trigger the GC process of the JVM.

After changing to query one by one, you can solve the problem of too many query results, but it will cause the method to cycle too many times, frequently query the DB, put too much pressure on the DB, and take too long time.

gong:
@fuweng11 Maybe add some description that why it can solve this fix high CPU usage.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
fuweng11: The previous query was to pull out all Streamsources in the templateId and filter them.
If there were 200 agents, this method would generate 500 + streamSource instances each time, leading to a large amount of gc.
Filtering is now done directly in the mysql layer to avoid generating too many instances and causing gc.

fuweng11:
> @fuweng11 Maybe add some description that why it can solve this fix high CPU usage.

done.

thexiay (approved):

thexiay:
LTGM

yunqingmoswu (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/core/impl/AgentServiceImpl.java
```diff
@@ -338,30 +336,24 @@ private void preProcessTemplateFileTask(TaskRequest taskRequest) {
...
-                    fileEntity.setTemplateId(sourceEntity.getId());
-                    fileEntity.setAgentIp(agentIp);
-                    fileEntity.setStatus(SourceStatus.TO_BE_ISSUED_ADD.getCode());
-                    // create new sub source task
-                    sourceMapper.insert(fileEntity);
-                    LOGGER.info("Transform new template task({}) for agent({}) in cluster({}).",
-                            fileEntity.getId(), taskRequest.getAgentIp(), taskRequest.getClusterName());
+        sourceEntities.forEach(sourceEntity -> {
+                    StreamSourceEntity subSource = sourceMapper.selectExistsByTemplateIdAndIp(sourceEntity.getId(),
+                            agentIp);
```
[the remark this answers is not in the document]
fuweng11: It won't.

dockerzhang (approved):

thexiay:
please apply `mvn spotless:apply`

healchow (approved):

gong (approved):

PR #7590, opened 2023-03-14
[INLONG-7589][Sort] Support multi node relation with same output but different input nodes

### LINKED ISSUE #7589: [Feature][Sort] Support multi node relation with same output but different input nodes
### Description

Now it's not capable for sort to generate multi insert sqls for one output with multiple inputs which is not acceptable for user cases. 

For instance, user may want to get data from multiple data sources, they can only use union operation for now. But union operation in flink is too heavy.

We need to support generate sqls from one stream such that:

Insert into sink select * from sourceA;
Insert into sink select * from sourceB; 

### Use case

We need to support generate sqls from one stream such that:

Insert into sink select * from sourceA;
Insert into sink select * from sourceB; 

### Code of Conduct

EMsnap:
### Prepare a Pull Request
- Fixes #7589 

### Motivation

Now it's not capable for sort to generate multi insert sqls for one output with multiple inputs which is not acceptable for user cases.

For instance, user may want to get data from multiple data sources, they can only use union operation for now. But union operation in flink is too heavy.

We need to support generate sqls from one stream such that:

Insert into sink select * from sourceA;
Insert into sink select * from sourceB;

### Modifications

Change the `parseNode` to `ParseSingleNode` because the recuisive method will be multi branches which is not suitable for reading. And move the input nodes parsing outside of `parseNode` method to make it easy for reading

### Verifying this change

Run `AllmigrateTestMultiRelations`

### DIFF
--- a/inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateTest.java
+++ b/inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateTest.java
@@ -66,6 +66,23 @@ public class AllMigrateTest {
                 ExtractMode.CDC, null, null);
     }
 
+    private MySqlExtractNode buildAllMigrateExtractNode2() {
+
+        Map<String, String> option = new HashMap<>();
+        option.put("append-mode", "false");
+        option.put("migrate-all", "true");
+        List<String> tables = new ArrayList(10);
+        tables.add("test.*");
+        List<FieldInfo> fields = Collections.singletonList(
+                new MetaFieldInfo("data", MetaField.DATA));
+
+        return new MySqlExtractNode("2", "mysql_input", fields,
+                null, option, null,
+                tables, "localhost", "root", "inlong",
+                "test", null, null, true, null,
+                ExtractMode.CDC, null, null);
+    }
+
     private MySqlExtractNode buildAllMigrateExtractNodeWithBytesFormat() {
         List<FieldInfo> fields = Collections.singletonList(
                 new MetaFieldInfo("data", MetaField.DATA_BYTES_DEBEZIUM));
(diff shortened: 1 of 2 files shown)


inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateTest.java
```diff
@@ -132,14 +149,43 @@ public void testAllMigrate() throws Exception {
...
+            .useBlinkPlanner()
+            .inStreamingMode()
+            .build();
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        env.setParallelism(1);
+        env.enableCheckpointing(10000);
+        StreamTableEnvironment tableEnv = StreamTableEnvironment.create(env, settings);
+        Node inputNode = buildAllMigrateExtractNode();
+        Node inputNode2 = buildAllMigrateExtractNode2();
+        Node outputNode = buildAllMigrateKafkaNode();
```
gong, line 178: kafka node id need change to 3

gong: UT generated sql maybe error but don't throw exception. It can't check sql is right statement.

gong: you can check sql 
image

EMsnap: ut has been fixed thx

e-mhui (approved):

healchow (approved):

dockerzhang (approved):

gong (approved):

PR #7864, opened 2023-04-16
[INLONG-6668][Sort] Reusing existing constants

### LINKED ISSUE #6668: [Improve][Sort] Unify the constants in `inlong-common`  and `sort-connectors`
### Description

many constants defined in both, it's redundant, it's better to unify them, for example `metrics.audit.proxy.hosts`:

- `sort-connectors`
image

- `inlong-common`
image

### InLong Component

InLong Sort

### Code of Conduct

bibhu107:
### Prepare a Pull Request

- Fixes #6668 

### Motivation

Using already existing constants and removing hard codings

### Documentation

  - Does this pull request introduce a new feature? no
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)  not documented
  - If a feature is not applicable for documentation, explain why? This change needs no documentation as it's using already created constants at hard-coded strings 

### DIFF
--- a/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
+++ b/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
@@ -40,7 +40,6 @@ public class Constants {
 
     public static final String DRAIN = "flink.drain";
 
-    public static final String METRICS_AUDIT_PROXY_HOSTS_KEY = "metrics.audit.proxy.hosts";
 
     // dataflow
     public static final String SOURCE_INFO = "source_info";
--- a/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
+++ b/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
@@ -40,7 +40,6 @@ public class Constants {
 
     public static final String DRAIN = "flink.drain";
 
-
     // dataflow
     public static final String SOURCE_INFO = "source_info";
 
--- a/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
+++ b/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
@@ -40,7 +40,7 @@ public class Constants {
 
     public static final String DRAIN = "flink.drain";
 
-    public static final String METRICS_AUDIT_PROXY_HOSTS = "metrics.audit.proxy.hosts";
+    public static final String METRICS_AUDIT_PROXY_HOSTS_KEY = "metrics.audit.proxy.hosts";
 
     // dataflow
     public static final String SOURCE_INFO = "source_info";
@@ -79,9 +79,9 @@ public class Constants {
     public static final String SEPARATOR = ":";
 
     /**
-     * Generate the Job name through {@link ProcessForm}: <br/> 
-     * When the ProcessForm is {@link GroupResourceProcessForm}, the format of the job name is 'InLong-Sort-{Group ID}', 
-     * otherwise take the  {@link Constants#DEFAULT_SORT_JOB_NAME}: 'InLong-Sort-Job'. 
+     * Generate the Job name through {@link ProcessForm}: <br/>
+     * When the ProcessForm is {@link GroupResourceProcessForm}, the format of the job name is 'InLong-Sort-{Group ID}',
+     * otherwise take the  {@link Constants#DEFAULT_SORT_JOB_NAME}: 'InLong-Sort-Job'.
      */
     public static Function<ProcessForm, String> SORT_JOB_NAME_GENERATOR =
             (ProcessForm processForm) -> Optional.of(processForm)
--- a/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
+++ b/inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
@@ -29,7 +29,7 @@ import java.util.Properties;
 
 import static org.apache.inlong.manager.plugin.flink.enums.Constants.ADDRESS;
 import static org.apache.inlong.manager.plugin.flink.enums.Constants.JOB_MANAGER_PORT;
-import static org.apache.inlong.manager.plugin.flink.enums.Constants.METRICS_AUDIT_PROXY_HOSTS;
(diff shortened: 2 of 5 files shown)


dockerzhang:
@bibhu107 You could run `mvn spotless:apply` to fix the check style error.

inlong-sort/sort-connectors/base/src/main/java/org/apache/inlong/sort/base/Constants.java
```diff
@@ -20,6 +20,8 @@
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.ConfigOptions;
 import org.apache.inlong.sort.base.sink.SchemaUpdateExceptionPolicy;
+import static org.apache.inlong.manager.plugin.flink.enums.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY;
```
healchow, line 23: Hi, the `inlong-sort` module should not dependency the `inlong-manager` module directly, can you move the `org.apache.inlong.manager.plugin.flink.enums.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY` to `inlong-common` module?

bibhu107: Hi @healchow 
Have taken  care of this in next commit. Please let me know if the changes look good?

bibhu107:
> @bibhu107 You could run `mvn spotless:apply` to fix the check style error.

With second commit ran `mvn spotless:apply`, it built successfully.

EMsnap (approved):

dockerzhang (approved):

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
```diff
@@ -29,7 +29,7 @@
 
 import static org.apache.inlong.manager.plugin.flink.enums.Constants.ADDRESS;
 import static org.apache.inlong.manager.plugin.flink.enums.Constants.JOB_MANAGER_PORT;
-import static org.apache.inlong.manager.plugin.flink.enums.Constants.METRICS_AUDIT_PROXY_HOSTS;
+import static org.apache.inlong.manager.plugin.flink.enums.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY;
```
gong, line 32: Why don't you use `org.apache.inlong.common.constant.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY`?

healchow (approved):

inlong-sort/sort-connectors/base/src/main/java/org/apache/inlong/sort/base/Constants.java
healchow: LGTM

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
bibhu107: Hi @gong 
I tried adding `org.apache.inlong.common.constant.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY` to `org.apache.inlong.common.constant.Constants` and also directly using in `inlong-sort` but in both the options were needing to create dependency on 'manager-plugins' which is not correct which might create  circular dependency between modules 'inlong-common' and 'manager-common'.
For reference you can check commit-1

yunqingmoswu (approved):

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
gong: > Hi @gong I tried adding `org.apache.inlong.common.constant.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY` to `org.apache.inlong.common.constant.Constants` and also directly using in `inlong-sort` but in both the options were needing to create dependency on 'manager-plugins' which is not correct which might create circular dependency between modules 'inlong-common' and 'manager-common'. For reference you can check commit-1

@bibhu107 Hi, I think you can `inlong-sort` and `manager-plugin`  module dependency `inlong-common`  after adding `org.apache.inlong.common.constant.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY` to `org.apache.inlong.common.constant.Constants`. `inlong-sort` module don't need to dependency `manager-plugin` module.

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
```diff
@@ -40,7 +40,7 @@ public class Constants {
 
     public static final String DRAIN = "flink.drain";
 
-    public static final String METRICS_AUDIT_PROXY_HOSTS = "metrics.audit.proxy.hosts";
+    public static final String METRICS_AUDIT_PROXY_HOSTS_KEY = "metrics.audit.proxy.hosts";
```
healchow, line 43: Can we remove this constant, and using the same constant from the `inlong-common` module?
I verified that the `manager-plugins` module has already dependent the `inlong-common` module.

healchow (changes requested):

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
bibhu107: Hi @healchow 
Thanks for the suggestions. Have taken care of these in new commit. Please check

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
bibhu107: Hi @gong 

Have considered the above comments and raised a new commit. Please check and let me know if it looks good.

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/flink/enums/Constants.java
healchow: LGTM

gong (approved):

inlong-manager/manager-plugins/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
gong: > Hi @gong
> 
> Have considered the above comments and raised a new commit. Please check and let me know if it looks good.

look good to me

PR #7924, opened 2023-04-26
[INLONG-7900][Sort] Support partition by primary key when upsert single table of Kafka

### LINKED ISSUE #7900: [Feature][Sort] Support partition by custom fields when upsert single table of Kafka
### Description

Currently, Kafka sink single table partition strategy is defaulted to round-robin.
This will cause binlogs to be in different partitions, unordered. 
As a result, some Binlogs will be lost when consuming kafka.

TODO:
Kafka single table support partition by primary key

### Use case
Run the following code:

        String insertQuery = "INSERT INTO test1(id) VALUES(?)";
        PreparedStatement preparedStmt = conn.prepareStatement(insertQuery);
        for(int i=1;i<=5000;i++){
            preparedStmt.setInt(1, i);
            preparedStmt.addBatch();
        }
        preparedStmt.executeBatch();

        String upsertQuery = "update test1 SET id=? WHERE id=? ";
        PreparedStatement preparedStmt2 = conn.prepareStatement(upsertQuery);
        for(int i=1;i<=5000;i++){
            preparedStmt2.setInt(1, i+5000);
            preparedStmt2.setInt(2, i);
            preparedStmt2.addBatch();
        }
        preparedStmt2.executeBatch();

        PreparedStatement preparedStmt3 = conn.prepareStatement(upsertQuery);
        for(int i=1;i<=5000;i++){
            preparedStmt3.setInt(1, i+10000);
            preparedStmt3.setInt(2, i+5000);
            preparedStmt3.addBatch();
        }
        preparedStmt3.executeBatch();

        */
        String deleteQuery = "DELETE FROM test1 WHERE id=?";
        PreparedStatement preparedStmt3 = conn.prepareStatement(deleteQuery);
        for(int i=10001;i<=15000;i++){
            preparedStmt3.setInt(1, i);
            preparedStmt3.addBatch();
        }
        preparedStmt3.executeBatch();

from mysql2kafka,there should be 20000 messages.
from kafka2mysql, there should be 0 entries, since everything is deleted. However, actually, 
image
3892 messages remain.

### Code of Conduct

Yizhou-Yang:
### Prepare a Pull Request
- Fixes #7900 

### Motivation
See issue.

### Modifications
Added PrimaryKeyPartitioner, and made modifications to support upsert-kafka

### Verifying this change
1. execute the sql in the issue again
2. create a kafka-mysql task to consume all the messages back to mysql

the mysql table fluctuated from 1000-2000 rows, and in the end it reduced to 0 rows, so I assume that the issue has been solved, since if it is still round robin, some update queries will be unordered, and it is unlikely that 3890 remaining records can be reduced to 0. The format used is json, for canal-json/avro/csv, I only tested that the primaryKey partitition is in effect and there are no errors.

### DIFF
--- a/inlong-sort/sort-connectors/kafka/src/main/java/org/apache/inlong/sort/kafka/partitioner/PrimaryKeyPartitioner.java
+++ b/inlong-sort/sort-connectors/kafka/src/main/java/org/apache/inlong/sort/kafka/partitioner/PrimaryKeyPartitioner.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sort.kafka.partitioner;
+
+import org.apache.flink.streaming.connectors.kafka.partitioner.FlinkKafkaPartitioner;
+import org.apache.flink.util.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The PrimaryKey Partitioner is used to extract primary key from single table messages:
+ *
+ * @param <T>
+ */
+public class PrimaryKeyPartitioner<T> extends FlinkKafkaPartitioner<T> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(RawDataHashPartitioner.class);
+    private static final long serialVersionUID = 1L;
+
+    @Override
+    public void open(int parallelInstanceId, int parallelInstances) {
+        super.open(parallelInstanceId, parallelInstances);
+    }
+
+    @Override
+    @SuppressWarnings({"unchecked"})
+    public int partition(T record, byte[] key, byte[] value, String targetTopic, int[] partitions) {
+        Preconditions.checkArgument(
+                partitions != null && partitions.length > 0,
+                "Partitions of the target topic is empty.");
+        try {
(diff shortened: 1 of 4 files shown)


inlong-sort/sort-connectors/kafka/src/main/java/org/apache/inlong/sort/kafka/partitioner/PrimaryKeyPartitioner.java
duowan1520, line 1: Hi, I noticed that in the RawDataHashPartitioner class there is code logic to get partition keys based on raw data primarykey and pattern. We have added a new partitioner called PrimaryKeyPartitioner and I was wondering what the relationship is between PrimaryKeyPartitioner and RawDataHashPartitioner?

Yizhou-Yang: Hi! The rawdatahashpartitioner was based on RawData, which are serialized canal/debezium data and are deserialized using corresponding JsonDynamicFormat. However, the PrimaryKeyPartitioner need to support csv/avro/json format in addition to canal/debezium, which does not have JsonDynamicFormat and need not to be parsed. If we use RawDataHashPartitioner, there will be Null Pointer Exceptions from the missing JsonDynamicFormat.

Yizhou-Yang: updating this pr, the user should designate the "partition key" using an additional property, and this new entry will be parsed from original data and used for hashing.

duowan1520: Should we consider elevating the logic for extracting primary keys to a higher level and then merging the PrimaryKeyPartitioner and RawDataHashPartitioner into a hash-based Partitioner? This would help avoid introducing different partitioners when choosing between synchronizing a single table or multiple tables.

Yizhou-Yang: I adjusted the implementation to make it schema-reliant, so that it directly extracts from rowdata instead of the serialized key now. please refresh the code to see the new implementation.. I think it is pretty distinct from raw hash partitioner now, and might take some effort to merge.

The two scenerios are pretty different, when trying merge the two scenerios together ( like how i tried in Doris connector), the resulting code has proven to be.. rather hard to debug and modify, since all changes need to be compatible with both scenerios. I think that segregating the two cases strictly within connectors might lead to fewer bugs, especially in the production environment.

PR #7942, opened 2023-04-27
[INLONG-7931][DataProxy] Optimize common.properties related control mechanism

### LINKED ISSUE #7931: [Improve][DataProxy] Optimize common.properties related control mechanism
### Description

In the current configuration management of DataProxy, there are two parts related to common.properties control implementation, CommonPropertiesHolder and ConfigManager.getCommonProperties(). According to the analysis, this part needs to simplify the two sets of implementations into one set, and at the same time, the relevant implementation logic is adjusted as follows :

1. It is necessary to add a mandatory requirement for common.properties configuration: the configuration file must exist when the node starts, if the file does not exist, the file fails to be read, the file content parsing fails, or the required fields in the file are not configured, the DataProxy node must output the error log, and exits;
2. common.properties is only loaded at startup, and does not need to be loaded during operation;
3. The acquisition of the Manager IP address in common.properties must support multiple strategies, such as configuring the domain name port, configuring the IP and port, or obtaining the corresponding Manager connection address provided by other third-party components;
4. Unify the field configuration tags in common.properties, and align the field configuration tags that are repeated, have the same meaning but misaligned names, and are scattered everywhere

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #7931

Modification details:
1. Merge CommonPropertiesHolder and ConfigManager.getCommonProperties() into CommonConfigHolder, manage configuration logic related to common.properties, and replace related content;
2. Converge the field key defined in common.properties, and the default value logic, and unify the value logic of the field;
3. Optimize some common.properties related implementation codes and remove unnecessary classes;


inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
```diff
@@ -0,0 +1,477 @@
...
+    public String getProxyNodeId() {
+        return proxyNodeId;
+    }
+
+    public String getMsgCompressType() {
+        return msgCompressType;
+    }
+
+    private void preReadFields() {
+        String tmpValue;
```
healchow, line 284: Can we use Properties or JSON parsing tools to automatically parse the configuration in the file, and then map it to a Java POJO?
Now this kind of modification is not very flexible, such as modifying or adding a new configuration, you also need to modify the code.

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
```diff
@@ -0,0 +1,477 @@
...
+        if (StringUtils.isBlank(fieldVal)) {
+            LOG.error("Required {} field value is blank in {}, exit!",
+                    requiredFieldKey, COMMON_CONFIG_FILE_NAME);
+            System.exit(5);
+        }
+    }
+
+    private boolean loadConfigFile() {
+        InputStream inStream = null;
+        try {
```
healchow, line 433: Using `try-with-resource` can reduce the `finally` statement.

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
```diff
@@ -0,0 +1,477 @@
...
+     * @param defaultValue
+     * @return
+     */
+    public static String getStringFromContext(Context context, String key, String defaultValue) {
+        String value = context.getString(key);
+        value = (value != null) ? value : getInstance().getProperties().getOrDefault(key, defaultValue);
+        return value;
+    }
+
+    public String getClusterTag() {
```
healchow, line 187: Too much getters and setters, can we use the `lombok` here to reduce them?

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
gosonzhang: It is the construction of the Properties parsing mode, and the pre-reading fields are optimized for efficiency

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
gosonzhang: This is just a coding habit and not a problem. Try-with-resource is just a syntactic sugar. Personally, I tend to explicitly clean up resources during the release phase.

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/config/CommonConfigHolder.java
gosonzhang: This part is pre-reading processing, used for speed-up, and the developed code cannot be reduced through lombok

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/metrics/DataProxyMetricItem.java
```diff
@@ -144,7 +144,8 @@ public static void fillInlongId(Event event, Map<String, String> dimensions) {
      */
     public static void fillAuditFormatTime(Event event, Map<String, String> dimensions) {
         long msgTime = (event != null) ? AuditUtils.getLogTime(event) : System.currentTimeMillis();
-        long auditFormatTime = msgTime - msgTime % CommonPropertiesHolder.getAuditFormatInterval();
+        long auditFormatTime =
```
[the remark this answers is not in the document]
gosonzhang: done

luchunliang (approved):
LGTM

baomingyu (approved):
+1

woofyzhao:
+1

PR #7960, opened 2023-05-06
[INLONG-7958][Sort] Fix MongoDB's schema becomes unordered after extracting the row data

### LINKED ISSUE #7958: [Bug][Sort] MongoDB's schema becomes unordered after extracting the row data
### What happened

MongoDB's schema becomes unordered after extracting the row data.

image

The schema of MongoDB maintains consistency during sync.

### How to reproduce

none.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Sort

### Code of Conduct

e-mhui:
### Prepare a Pull Request

[INLONG-7958][Sort] Fix MongoDB's schema becomes unordered after extracting the row data

- Fixes #7958 

### Motivation

1. MongoDB's schema becomes unordered after extracting the row data. The sink (e.g. iceberg) automatically builds the table, and the source schema being shuffled during the synchronization process will result in inconsistent schema order between source and sink.
image

2. For all database migration, Mognodb CDC does not specify a primary key, so **upsert** cannot be implemented. Refer to https://ververica.github.io/flink-cdc-connectors/master/content/connectors/mongodb-cdc.html, we should specify `_id` as the default primary key.

> MongoDB’s change event record doesn’t have updated before message. So, we can only convert it to Flink’s UPSERT changelog stream. An upsert stream requires a unique key, so we must declare _id as primary key. We can’t declare other column as primary key, because delete operation does not contain the key and value besides _id and sharding key.

### Modifications

1. Use `LinkedHashMap` instead of `HashMap` to make the schema ordered.
2. Use `_id` as the default primary key
3. Refactor the code

### DIFF
--- a/inlong-sort/sort-connectors/mongodb-cdc/src/main/java/org/apache/inlong/sort/cdc/mongodb/debezium/DebeziumJson.java
+++ b/inlong-sort/sort-connectors/mongodb-cdc/src/main/java/org/apache/inlong/sort/cdc/mongodb/debezium/DebeziumJson.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


EMsnap (approved):

yunqingmoswu (approved):

gong:
@e-mhui Why to keep schema order? I suggest that add reason.

e-mhui:
> @e-mhui Why to keep schema order? I suggest that add reason.

The sink (e.g. iceberg) automatically builds the table, and the source schema being shuffled during the synchronization process will result in inconsistent schema order between source and sink.

e-mhui:
> @e-mhui Why to keep schema order? I suggest that add reason.

The reason has been added to the description.

gong:
> > @e-mhui Why to keep schema order? I suggest that add reason.
> 
> The sink (e.g. iceberg) automatically builds the table, and the source schema being shuffled during the synchronization process will result in inconsistent schema order between source and sink.

look good to me

gong (approved):

dockerzhang (approved):

PR #8015, opened 2023-05-11
[INLONG-8013][Sort] Fix UT failed caused by OOM

### LINKED ISSUE #8013: [Bug][Sort] Cause lots of OOM UT failed
### What happened

Cause logs of OOM UT failed.

image

All UTs passed

### How to reproduce

see the latest workflow.

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Sort

### Code of Conduct

e-mhui:
### Prepare a Pull Request

[INLONG-8013][Sort] Fix UT failed

- Fixes #8013 

### Motivation

There are two ways to solve it:

1. All test classes need to inherit from `AbstractTestBase`, because `AbstractTestBase` can reuse the same Flink cluster, which can reduce resource usage.

image

2. Add the `taskmanager.memory.network.min` parameter to limit the memory consumption of the network (the default memory for the network is 64 MB). More details see: https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/deployment/memory/mem_setup_tm/

```java
        EnvironmentSettings settings = EnvironmentSettings
                .newInstance()
                .useBlinkPlanner()
                .inStreamingMode()
                .build();
        Configuration configuration = new Configuration();
        // Set `taskmanager.memory.network.min` to avoid NetworkBufferPool OOM during run workflow.
        configuration.setString("taskmanager.memory.network.min", "1 mb");
        env = StreamExecutionEnvironment.getExecutionEnvironment(configuration);
        env.setParallelism(1);
        env.enableCheckpointing(10000);
        this.tableEnv = StreamTableEnvironment.create(env, settings);
```

### DIFF
--- a/inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateMongoDBTest.java
+++ b/inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateMongoDBTest.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sort.parser;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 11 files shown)


inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateMongoDBTest.java
```diff
@@ -40,6 +40,7 @@
 import org.apache.inlong.sort.protocol.node.load.KafkaLoadNode;
 import org.apache.inlong.sort.protocol.transformation.FieldRelation;
 import org.apache.inlong.sort.protocol.transformation.relation.NodeRelation;
+import org.junit.Assert;
```
healchow, line 43: Why import this class here?

EMsnap:
maybe this is not the reason for the failure of UT case and we should find out the root cause of the OOM. Maybe we can login to the compile machine to analyse  ?

yunqingmoswu (approved):

inlong-sort/sort-core/src/test/java/org/apache/inlong/sort/parser/AllMigrateMongoDBTest.java
e-mhui: This PR  https://github.com/apache/inlong/pull/7998 caused OOM when running the workflow. Adding this is to test whether there is any impact.

dockerzhang (approved):

healchow (approved):

PR #8041, opened 2023-05-17
[INLONG-7950][DataProxy] Optimize the implementation logic of the Source

### LINKED ISSUE #7950: [Improve][DataProxy] Optimize the implementation logic of the Source
### Description

At present, there are several problems in the Source:
1. InLongMsg V0 and V1 versions use two similar implementations for processing, and here we consider merging them for easy maintenance and management;
2. Consider extracting the common implementations of HTTP, TCP, and UDP on the source side, if there are other different encoding schemes, you only need to implement the corresponding MessageFactory and MessageHandler, and the related protocol Source processing does not need to be changed, which is convenient for other protocol extensions;
3. Adjust relevant implementation code, optimize logic, and supplement test cases

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #7950

### Motivation

1. Optimize the implementation logic of BaseSource, SimpleTcpSource, and SimpleUdpSource on the Source side, and align the setting parameters of the context configuration;
2. Optimize the message processing Handler of InLongV0Msg and InLongV1Msg on the Source side, and unify the processing framework;
3. Optimize InLongV0Msg decoding and packaging logic
4. Adjust the transfer parameters of SinkRspEvent, and adjust ChannelHandlerContext to Channel parameters


inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/source2/InLongMessageFactory.java
```diff
@@ -0,0 +1,77 @@
...
+        if (source.getChannelProcessor() != null) {
+            try {
+                Class<? extends ChannelInboundHandlerAdapter> clazz =
+                        (Class<? extends ChannelInboundHandlerAdapter>) Class.forName(source.getMessageHandlerName());
+                Constructor<?> ctor = clazz.getConstructor(BaseSource.class);
+                ChannelInboundHandlerAdapter messageHandler =
+                        (ChannelInboundHandlerAdapter) ctor.newInstance(source);
+                ch.pipeline().addLast("messageHandler", messageHandler);
+            } catch (Exception e) {
+                LOG.info("{} newInstance {} failure!", source.getName(),
```
healchow, line 72: Sorry, but why not `error` level here?

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/source2/SourceConstants.java
```diff
@@ -0,0 +1,202 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.dataproxy.source2;
+
+public class SourceConstants {
+
+    // source host
+    public static final String SRCCXT_CONFIG_HOST = "host";
```
healchow, line 23: I see a lot of the same constants, can we define common constants in the inlong-common module to improve code reuse?

gosonzhang: Well, there are indeed some places that will be repeated, but this is a temporary situation. After source2 is stabilized, the contents of the source and http directories will be replaced in parallel. The current modification idea is to modify them in parallel first, and finally remove them completely. Unused tags , the definition will not be retained later

inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/source2/InLongMessageFactory.java
gosonzhang: Done

luchunliang (approved):
LGTM

dockerzhang (approved):

dockerzhang (approved):

dockerzhang (approved):

PR #8053, opened 2023-05-18
[INLONG-8046][Dashboard] Support batch import of sink fields

### LINKED ISSUE #8046: [Improve][Dashboard] Support batch import of sink fields
### Description

_No response_

### InLong Component

InLong Dashboard

### Code of Conduct

featzhang:
- Fixes #8046 

## Append field names if absent
QQ20230519-125903-HD(1)

## Overwrite field names

overwrite-field-by-batch-import(1)

### DIFF
--- a/inlong-dashboard/src/ui/components/EditableTable/index.tsx
+++ b/inlong-dashboard/src/ui/components/EditableTable/index.tsx
@@ -69,6 +69,10 @@ export interface EditableTableProps
   // Can add a new line? Default: true.
   canAdd?: boolean;
   canBatchAdd?: boolean;
+  upsetByFieldKey?: boolean;
+  fieldNameKey?: string;
+  fieldTypeKey?: string;
+  fieldCommentKey?: string;
 }
 
 const getRowInitialValue = (columns: EditableTableProps['columns']) =>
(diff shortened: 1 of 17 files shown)


healchow (approved):

dockerzhang (approved):

inlong-dashboard/src/ui/components/EditableTable/index.tsx
```diff
@@ -69,6 +69,10 @@ export interface EditableTableProps
   // Can add a new line? Default: true.
   canAdd?: boolean;
   canBatchAdd?: boolean;
```
leezng, line 71: `EditableTable` as a base component, I don't think `upsetByFieldKey`, `fieldNameKey`, ... such parameters with obvious meanings. The parameters of `EditableTable` should be abstracted, which is the common capability of the underlying component, and the parameters added above should be implemented in the upper component.

For example, this PR adds a new capability to EditableTable, we can define it as an attribute related to append or overwrite, and the impact on fieldName and fieldKey should be implemented in the upper-level business component of Sink.

leezng (changes requested):

inlong-dashboard/src/ui/components/EditableTable/index.tsx
leezng: A key point here is that the general abstraction of the parameters of the underlying components and the concretization of the parameters of the upper components.

fuweng11 (approved):

healchow (approved):

leezng (approved):

PR #8145, opened 2023-06-01
[INLONG-8142][DataProxy] Optimize ConfigManager implementation (Part three)

### LINKED ISSUE #8142: [Improve][DataProxy] Optimize ConfigManager implementation (Part three)
### Description

Adjust the operation mode of topic metadata and mq cluster information, currently there are two ways to obtain the mapping relationship between groupId/streamId and topic, and metadata information such as MQ cluster: getConfig and getAllConfig, these two methods save metadata in different , and use different logic to interact with the Manager to get the latest configuration.

Through the analysis of the two methods, the design of getAllConfig is more in line with the requirements of online use. Only the metadata update request response will carry the full amount of metadata, at the same time, the response structure is easier to expand, but the configuration loading mode is not friendly to use: the metadata configuration is carried through the dataproxy.conf file, this kind of different configurations is mixed together, which is not convenient for system maintenance. The problem with the getConfig method is that the scalability of the response structure is not good enough, every request will return the full amount of metadata, regardless of whether the metadata changes.

Adjust this part in this issue: use getAllConfig as the only metadata configuration acquisition method, save metadata in a separate configuration file, adjust the acquisition and processing logic of Topic and MQ cluster information, and remove the implementation logic of the original getConfig and getAllConfig methods

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #8142 

1. Add MetaConfigHolder class to manage metadata information of Topic and MQ cluster;
2. Update the ConfigManager class, replace getConfig with getAllConfig, and adjust the related metadata update logic and query API;
3. Remove the metadata information management classes and interfaces of the existing Topic and MQ clusters;
4. Update the metadata change synchronization logic of the MQ cluster on the Sink side
5. Other small changes such as log output and file location migration


luchunliang (approved):
LGTM

aloyszhang (approved):

PR #8149, opened 2023-06-02
[INLONG-8148][Manager] Support precise query of cluster nodes based on IP

### LINKED ISSUE #8148: [Bug][Manager] The method of querying cluster nodes is not idempotent
### What happened

as we see in the picture, one agent is '10.19.132.152' and next agent is '10.19.132.15'
use selectByIpAndCluster , two agents will be selected in turn because modify_time is always change and select method use like instead of '='
image

the method must be idempotence

### How to reproduce

use local database will reproduce this scene

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

kipshi:
### Prepare a Pull Request

- Fixes #8148 

### Modifications

use '='  instead of 'like' 

### Verifying this change

use local database can verify

### Documentation

### DIFF
--- a/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterNodeEntityMapper.xml
+++ b/inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterNodeEntityMapper.xml
@@ -114,8 +114,8 @@
             </if>
             <if test="keyword != null and keyword != ''">
                 and (
-                ip like CONCAT('%', #{keyword}, '%')
-                or port like CONCAT('%', #{keyword}, '%')
+                ip  = #{keyword, jdbcType=VARCHAR}
+                or port = #{keyword, jdbcType=VARCHAR}
                 )
             </if>
         </where>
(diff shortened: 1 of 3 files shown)


inlong-manager/manager-dao/src/main/resources/mappers/InlongClusterNodeEntityMapper.xml
```diff
@@ -114,8 +114,8 @@
             </if>
             <if test="keyword != null and keyword != ''">
                 and (
-                ip like CONCAT('%', #{keyword}, '%')
```
fuweng11, line 117: keyword is used for fuzzy queries.For an exact lookup, add
```
 <if test="ip != null and ip != ''">
      and ip = #{ip, jdbcType=VARCHAR}
 </if>
```

gong: > keyword is used for fuzzy queries.For an exact lookup, add
> 
> ```
>  <if test="ip != null and ip != ''">
>       and ip = #{ip, jdbcType=VARCHAR}
>  </if>
> ```

@fuweng11 It will cause agent exception when use like. So we should use =?

fuweng11: Exact lookup cannot be used in keyword.

fuweng11 (approved):

healchow (approved):

dockerzhang (approved):

PR #8187, opened 2023-06-07
[INLONG-8161][DataProxy] Optimize BatchPackProfile related classes implementation

### LINKED ISSUE #8161: [Improve][DataProxy] Optimize BatchPackProfile related classes implementation
### Description

In the current sink implementation, the event is obtained from the channel and is ready to be sent. V1 messages need to be sent in batches but V0 messages do not. Sink uses BatchPackProfile and its subclasses SimpleBatchPackProfileV0 and OrderBatchPackProfileV0 for corresponding processing.

From the analysis of the implementation, this class can be optimized, by defining an abstract class, the common attributes and operations of the three classes, BatchPackProfile, SimpleBatchPackProfileV0, and OrderBatchPackProfileV0, are extracted, and then the events of V0 and V1 are processed separately, which can simplify the current implementation.

This issue is for this category improvement

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #8161

1. Add the PackProfile base class and BatchPackProfile and SimplePackProfile subclasses, replace the original BatchPackProfile base class and OrderBatchPackProfileV0, SimpleBatchPackProfileV0 subclasses, and adjust the relevant message processing logic;
2. Add a minimum value check to the meta.config.sync.interval.ms configuration item
3. Add configuration item node.startup.via.meta.config.file.enable to control whether to download metadata from Manager


inlong-dataproxy/dataproxy-source/src/main/java/org/apache/inlong/dataproxy/sink/mq/BatchPackProfile.java
```diff
@@ -63,36 +54,17 @@ public BatchPackProfile(String uid, String inlongGroupId, String inlongStreamId,
...
      * @return
      */
-    public boolean addEvent(ProxyEvent event, long maxPackCount, long maxPackSize) {
+    public boolean addEvent(Event event, long maxPackCount, long maxPackSize) {
         long eventLength = event.getBody().length;
         if (count >= maxPackCount || (count > 0 && size + eventLength > maxPackSize)) {
             return false;
         }
-        this.events.add(event);
+        this.events.add((ProxyEvent) event);
```
baomingyu, line 62: Suggest events  type from List<ProxyEvent> events  to  List<Event> events

gosonzhang: The call side of this API is passed event by event. It essentially wants to accumulate a certain number and size before sending it. For this PR，only the type change is made, and then the logical level change is made later.

inlong-dataproxy/dataproxy-source/src/test/resources/common.properties
```diff
@@ -20,6 +20,8 @@ metricDomains=DataProxy
 metricDomains.DataProxy.domainListeners=org.apache.inlong.dataproxy.metrics.prometheus.PrometheusMetricListener
 metricDomains.DataProxy.snapshotInterval=60000
 
+node.startup.via.meta.config.file.enable=true
```
dockerzhang, line 23: Please add more description for this param.
otherwise, could we use `startup.using.local.meta.file.enable` to make this param easy to understand?

gosonzhang: It's really long, I was thinking about it when I first named it :)

dockerzhang (approved):

inlong-dataproxy/conf/common.properties
```diff
@@ -28,6 +28,8 @@ proxy.cluster.extTag=default=true
 proxy.cluster.inCharges=admin
 # synchronize interval of meta config (millisecond)
 meta.config.sync.interval.ms=10000
+# whether to startup using the local metadata.json file without connecting to the Manager
+startup.using.local.meta.file.enable=false
```
healchow, line 32: It does not need to be divided into too many levels, it can be similar to other config items, using the naming method of small camel case.

Because in common practice, parameter classification generally means that there are other parameters at the same level, such as the above `proxy.cluster` parameter, the reason for classifying the `cluster` is that it also has sub-level parameters such as `name`, `tag`, and `extTag`, etc.

For example, `startup.localMetaFile.enable` may also be enough in this scenario. It clearly indicates whether to enable local meta files at startup.

gosonzhang: Previously, parameter names were defined in various styles, but now all new parameters are in all lowercase, dotted, easy-to-understand mode, and the old parameters remain unchanged for the time being

luchunliang (approved):
LGTM

Issue #8091, opened 2023-05-25
[Bug][Manager] Unsupported FieldType is reported when the source and sink are both PostgreSQL


JeffWang79:
### What happened

字段存在 int4类型，映射字段INTEGER，源和目标都是postgres,报错Unsupported FieldType : INTEGER，1.7和1.6都有这问题。
![5025942cb231469028d63d2ee998112](https://github.com/apache/inlong/assets/90167565/b57e6928-abc5-46a9-952d-2294e5fec67c)
![70e6752bd6f92abe7a56c60d11f2231](https://github.com/apache/inlong/assets/90167565/cb82352a-d4a3-4ac9-815c-4b1e0e9a5755)

### What you expected to happen

![image](https://github.com/apache/inlong/assets/90167565/e0a6188b-8ff8-4493-878d-32358538c8d5)

### How to reproduce

![image](https://github.com/apache/inlong/assets/90167565/542484c3-de5b-475a-9a1a-c4af1883667f)

### Environment

centos7.6，docker一键部署

### InLong version

1.7.0

### InLong Component

InLong Manager

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

chestnut-c:
Ok thanks for the feedback, we will improve the data type mapping for various data sources

dockerzhang:
@JeffWang79, please use English to describe this issue. Thanks.

dockerzhang:
fixed by https://github.com/apache/inlong/pull/8223, close it.

PR #8223, opened 2023-06-13
[INLONG-8222][Manager] Support different data types mapping for different data sources by the strategy pattern

### LINKED ISSUE #8222: [Feature][Manager] Support different data types mapping for different data sources by the strategy pattern
### Description

Using strategist mode to support the mapping relationship between different data source custom data types and flink data types, 
It supports the mapping of different data types from different data sources as a whole, supplementing the deficiencies of the current types and fix bug ( https://github.com/apache/inlong/issues/8091 ).

### Use case

_No response_

### Code of Conduct

chestnut-c:
[INLONG-8222][Manager] Adopt strategist mode to support different types of data type mapping for different data sources.

- Fixes #8222

### Motivation

Using strategist mode to support the mapping relationship between different data source custom data types and inlong data types , It supports the mapping of different data types from different data sources as a whole, supplementing the deficiencies of the current types and fix bug ( https://github.com/apache/inlong/issues/8091 ).

### DIFF
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/DataTypeMappingStrategyEnum.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/DataTypeMappingStrategyEnum.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.manager.common.datatype;
+
+import org.apache.inlong.manager.common.datatype.datasource.PostgreSQLDataType;
+
+/**
+ * The enum of data type mapping strategies for different data sources.
+ */
+public enum DataTypeMappingStrategyEnum implements DataTypeMappingStrategy {
+
+    /**
+     * postgresql data type mapping operation strategy
+     */
+    POSTGRESQL_DATA_TYPE_STRATEGY {
+
+        @Override
+        public String getMappingDataType(String originalType) {
+            return PostgreSQLDataType.getMappingDataType(originalType);
+        }
+    },
+    ;
+
+    @Override
+    public abstract String getMappingDataType(String originalType);
+}
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
@@ -30,7 +30,7 @@ public interface BaseDataType {
     String getOriginalType();
 
     /**
-     * Get the target data type of flink data type mapping
+     * Get the target data type of inlong data type mapping
      *
      * @return The target data type
      */
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.manager.common.datatype.datasource;
+
+/**
+ * The interface of base data type mapping
+ */
+public interface BaseDataType {
+
+    /**
+     * Get the original data type.
+     *
+     * @return The original data type
+     */
+    String getOriginalType();
+
+    /**
+     * Get the target data type of flink data type mapping
+     *
+     * @return The target data type
+     */
+    String getTargetType();
+}
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
@@ -23,11 +23,11 @@ package org.apache.inlong.manager.common.fieldtype.datasource;
 public interface BaseFieldTypeMapping {
 
     /**
-     * Get the original field type.
+     * Get the source field type
      *
-     * @return The original field type
+     * @return The source field type
      */
-    String getOriginalType();
+    String getSourceType();
 
     /**
      * Get the target field type of inlong field type mapping
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
@@ -20,7 +20,7 @@ package org.apache.inlong.manager.common.fieldtype.datasource;
 /**
  * The interface of base field type mapping
  */
-public interface BaseFieldType {
+public interface BaseFieldTypeMapping {
 
     /**
      * Get the original field type.
--- a/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
+++ b/inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
@@ -19,6 +19,8 @@ package org.apache.inlong.manager.common.fieldtype.datasource;
 
 import org.apache.commons.lang3.StringUtils;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.stream.Stream;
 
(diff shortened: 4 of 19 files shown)


chestnut-c:
> @chestnut-c It need add type precision support. Only add simple type mapping is not enough. You can see `org.apache.inlong.manager.pojo.sort.util.FieldFormatUtils` to implement this function.


Make the basic type first to ensure that each source can run normally, and then further supplement the accuracy of the type.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
```diff
@@ -0,0 +1,38 @@
...
+     * Get the original data type.
+     *
+     * @return The original data type
+     */
+    String getOriginalType();
+
+    /**
+     * Get the target data type of flink data type mapping
+     *
+     * @return The target data type
```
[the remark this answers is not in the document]
chestnut-c: > It should be inlong type. inlong will convert to flink type for sort.

ok, i will change it

gong:
> > @chestnut-c It need add type precision support. Only add simple type mapping is not enough. You can see `org.apache.inlong.manager.pojo.sort.util.FieldFormatUtils` to implement this function.
> 
> Make the basic type first to ensure that each source can run normally, and then further supplement the accuracy of the type.

ok

gong:
@chestnut-c It will improve INLONG usability when collaborating with long dashboard. Because many datasource have different type. Thanks for your contribution.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/DataTypeMappingStrategyEnum.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * The enum of data type mapping strategies for different data sources.
+ */
+public enum DataTypeMappingStrategyEnum implements DataTypeMappingStrategy {
+
+    /**
+     * postgresql data type mapping operation strategy
+     */
+    POSTGRESQL_DATA_TYPE_STRATEGY {
```
fuweng11, line 30: Why use Enum？Can we use a better method？

inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/sort/node/base/ExtractNodeProvider.java
```diff
@@ -60,8 +61,23 @@ public interface ExtractNodeProvider extends NodeProvider {
      */
     default List<FieldInfo> parseStreamFieldInfos(List<StreamField> streamFields, String nodeId) {
         // Filter constant fields
+        return parseStreamFieldInfos(streamFields, nodeId, null);
```
fuweng11, line 64: Why is the `dataTypeMappingStrategy` set to empty? I don't think you should do this if you want to use the original fieldtype. There should be a class to handle cases where the current type does not support mapping.

chestnut-c: > Why is the `dataTypeMappingStrategy` set to empty? I don't think you should do this if you want to use the original fieldtype. There should be a class to handle cases where the current type does not support mapping.

Overload the current method, I think both will work

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/DataTypeMappingStrategyEnum.java
chestnut-c: > Why use Enum？Can we use a better method？


It has been modified.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/datatype/datasource/BaseDataType.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.manager.common.datatype.datasource;
+
+/**
+ * The interface of base data type mapping
+ */
+public interface BaseDataType {
```
fuweng11, line 23: Maybe `BaseFieldType`.

inlong-manager/manager-pojo/src/main/java/org/apache/inlong/manager/pojo/sort/node/base/ExtractNodeProvider.java
```diff
@@ -60,8 +61,23 @@ public interface ExtractNodeProvider extends NodeProvider {
...
+
+    /**
+     * Parse StreamFieldInfos
+     *
+     * @param streamFields The stream fields
+     * @param nodeId The node id
+     * @param dataTypeMappingStrategy The data type mapping operation strategy
+     * @return FieldInfo list
+     */
+    default List<FieldInfo> parseStreamFieldInfos(List<StreamField> streamFields, String nodeId,
```
fuweng11, line 75: This method is redundant.Just rewrite `parseStreamFieldInfos(List<StreamField> streamFields, String nodeId)`

chestnut-c: > parseStreamFieldInfos

overloaded method

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
```diff
@@ -0,0 +1,38 @@
...
+ * The interface of base field type mapping
+ */
+public interface BaseFieldTypeMapping {
+
+    /**
+     * Get the original field type.
+     *
+     * @return The original field type
+     */
+    String getOriginalType();
```
healchow, line 30: Maybe `sourceType` is more clearly to mapping `targetType`.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Get the field type of inlong field type mapping by the original field type.
+     *
+     * @param originalType the original field type
+     * @return the target field type of inlong field type mapping
+     */
+    public static String getFieldTypeMapping(String originalType) {
+        String dataType = StringUtils.substringBefore(originalType, LEFT_BRACKET).toUpperCase();
+        return Stream.of(values()).filter(v -> v.getOriginalType().equals(dataType))
```
healchow, line 158: Is it possible to construct a Map and then `get` it directly when using it?

The current way is to loop through it every time it is used, which is too inefficient.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.stream.Stream;
+
+import static org.apache.inlong.manager.common.consts.InlongConstants.LEFT_BRACKET;
+
+/**
+ * The enum of PostgreSQL field type mapping.
+ */
+public enum PostgreSQLFieldTypeMapping implements BaseFieldTypeMapping {
```
healchow, line 29: The enumeration class is still not easy to expand.

Is it possible to add the type mapping to the configuration file, and then the code reads and parses it?

Of course, in this issue, you can use the enumeration method first, and then you can take the time to see how to optimize it through the configuration file.

inlong-manager/manager-pojo/src/test/java/org/apache/inlong/manager/pojo/sort/util/FieldInfoUtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.inlong.manager.common.fieldtype.strategy.PostgreSQLFieldTypeStrategy;
+import org.apache.inlong.manager.pojo.stream.StreamField;
+import org.apache.inlong.sort.formats.common.IntTypeInfo;
+import org.apache.inlong.sort.formats.common.TypeInfo;
+import org.apache.inlong.sort.protocol.FieldInfo;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class FieldInfoUtilsTest {
```
healchow, line 29: Add javadoc please.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/strategy/DefaultFieldTypeStrategy.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.inlong.manager.common.fieldtype.strategy;
+
+/**
+ * The default field type mapping strategy
+ */
+public class DefaultFieldTypeStrategy implements FieldTypeMappingStrategy {
+
+    @Override
+    public String getFieldTypeMapping(String originalType) {
```
healchow, line 26: `originalType` -> `sourceType`?

chestnut-c: > `originalType` -> `sourceType`?


almost the same meaning,i will change it.

inlong-manager/manager-pojo/src/test/java/org/apache/inlong/manager/pojo/sort/util/FieldInfoUtilsTest.java
chestnut-c: > Add javadoc please.

okk.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
chestnut-c: > Is it possible to construct a Map and then `get` it directly when using it?
> 
> The current way is to loop through it every time it is used, which is too inefficient.

it's ok

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/BaseFieldTypeMapping.java
chestnut-c: > Maybe `sourceType` is more clearly to mapping `targetType`.

almost the same meaning,i will change it.

inlong-manager/manager-common/src/main/java/org/apache/inlong/manager/common/fieldtype/datasource/PostgreSQLFieldTypeMapping.java
chestnut-c: > The enumeration class is still not easy to expand.
> 
> Is it possible to add the type mapping to the configuration file, and then the code reads and parses it?
> 
> Of course, in this issue, you can use the enumeration method first, and then you can take the time to see how to optimize it through the configuration file.

It is really better to add the configuration file, and then optimize it in the next PR

healchow (approved):
LGTM

fuweng11 (approved):

dockerzhang (approved):

gong (approved):

PR #8285, opened 2023-06-19
[INLONG-8284][DataProxy] Unify the message encoding definition of DataProxy

### LINKED ISSUE #8284: [Improve][DataProxy] Unify the message encoding definition of DataProxy
### Description

After DataProxy receives the data reported by the business, it will encode the data reported by the business in the InLongMsg format or the PB format defined by MessagePack, and encapsulate the data in the MQ message body. After receiving the message, the receiver needs to follow the corresponding Only by decoding the format can the content in the MQ message body be obtained correctly.

Currently, there are two encoding format definition keys in DataProxy, EventConstants.HEADER_KEY_VERSION and ConfigConstants.MSG_ENCODE_VER. From code analysis, these two keys are the results of two identical ideas but different implementations; from the time of submission, HEADER_KEY_VERSION is defined earlier, but its name is too general, and it is easy to refer to it unclearly, especially when the business is directly reported to MQ, there is a high probability of conflict or unclear reference; MSG_ENCODE_VER is submitted later, and the value is not comprehensive enough.

Based on the above problems, submit this modification here, carry out the merge processing of this part, and only keep one key. At the same time, for compatibility, this version will set HEADER_KEY_VERSION to an expired state, and it will be deleted in a subsequent version.

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #8284 

1. Define a new class DataProxyMsgEncVer, whose value completely covers EventConstants.HEADER_KEY_VERSION and is consistent;
2. Add parallel processing to the places involving EventConstants.HEADER_KEY_VERSION and ConfigConstants.MSG_ENCODE_VER;
3. Adjust the implementation of MessageDeserializer, and use the DataProxyMsgEncVer enumeration value for judgment;
4. Optimize the SimplePackProfile class event to add the attribute value set of MQ message, which only contains necessary fields;
5. Other small optimization measures

### DIFF
--- a/inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyMsgEncVer.java
+++ b/inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyMsgEncVer.java
@@ -15,14 +15,19 @@
  * limitations under the License.
  */
 
-package org.apache.inlong.dataproxy.utils;
+package org.apache.inlong.common.enums;
 
-public enum InLongMsgVer {
+/**
+ * Enumeration class of encoding format of data output from DataProxy to MQ
+ */
+public enum DataProxyMsgEncVer {
 
-    INLONG_V0(0, "V0", "The inlong-msg V0 format"),
-    INLONG_V1(1, "V1", "The inlong-msg V1 format");
+    MSG_ENCODE_VER_RAW(0, "Raw", "Raw message without any InLong format"),
+    MSG_ENCODE_VER_PB(1, "PB", "The PB MessagePack encode format"),
+    MSG_ENCODE_VER_INLONGMSG(2, "InLongMsg", "The InLongMsg encode format"),
+    MSG_ENCODE_VER_UNKNOWN(99, "Unknown", "Unknown encode format");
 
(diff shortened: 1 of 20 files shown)


vernedeng (approved):
LGTM

inlong-common/src/main/java/org/apache/inlong/common/enums/DataProxyMsgEncVer.java
```diff
@@ -15,14 +15,19 @@
...
  */
 
-package org.apache.inlong.dataproxy.utils;
+package org.apache.inlong.common.enums;
 
-public enum InLongMsgVer {
+/**
+ * Enumeration class of encoding format of data output from DataProxy to MQ
+ */
+public enum DataProxyMsgEncVer {
```
aloyszhang, line 23: It looks more like the type or format of msg encoding than the version.

gosonzhang: Yes, you're right.

aloyszhang (approved):

dockerzhang (approved):

PR #8295, opened 2023-06-20
[INLONG-8294][DataProxy] Optimize the log output in the Sink module

### LINKED ISSUE #8294: [Improve][DataProxy] Optimize the log output in the Sink module
### Description

There are several problems with the log output under the Sink module:
1. It is not necessary to set the scope of the log object Public, and it is more appropriate to define it as private or protected;
2. The content of the log output is not very clear, and it is impossible to locate which Sink object generated the content;
3. The operation encapsulation of some objects is not very good, it is better to use the function method;
4. Part of the code is not used and needs to be removed

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #8294

Modifications:
1. Adjust the scope of the log object in the class from Public to private or protected;
2. Modify the log output content, add the Sink object name, and additional prompt information to facilitate the location and search of problem logs
3. Adjust the reference and operation of the dispatchQueue object, concentrate the object in the MessageQueueZoneSink class and encapsulate it with functions to avoid missing operations;
4. Remove the unused reloadTimer implementation in the SinkContext class;
5. Adjust the variable settings of related classes and add parameter setting instructions according to the IDE prompts


luchunliang (approved):
LGTM

baomingyu (approved):

PR #8344, opened 2023-06-27
[INLONG-8343][Dashboard] Increase the length of the field for creating MySQL varchar types to 16383

### LINKED ISSUE #8343: [Improve][Dashboard] Increase the length of the field for creating MySQL varchar types to 16383
### Description

The current limit for creating MySQL varchar type fields on the Dashboard page is no more than 255, but the length of 255 fields may not be sufficient during application. Therefore, I suggest increasing the length limit to 16383, as the length of MySQL varchar is limited by character set encoding and line record length. Therefore, we use the minimum length of 16383 that meets these two conditions as the limit for creating table fields.

### InLong Component

InLong Dashboard

### Code of Conduct

xuesongxs:
- Fixes #8343 

The current limit for creating MySQL varchar type fields on the Dashboard page is no more than 255, but the length of 255 fields may not be sufficient during application. Therefore, I suggest increasing the length limit to 16383, as the length of MySQL varchar is limited by character set encoding and line record length. Therefore, we use the minimum length of 16383 that meets these two conditions as the limit for creating table fields.


inlong-dashboard/src/plugins/sinks/defaults/MySQL.ts
```diff
@@ -47,7 +47,7 @@ const fieldTypesConf = {
   TIME: () => '',
   DATETIME: () => '',
   CHAR: (m, d) => (1 <= m && m <= 255 ? '' : '1<=M<=255'),
-  VARCHAR: (m, d) => (1 <= m && m <= 255 ? '' : '1<=M<=255'),
+  VARCHAR: (m, d) => (1 <= m && m <= 16383 ? '' : '1<=M<=16383'),
```
healchow, line 50: It seems that the maximum length of MySQL's varchar type is 65535?

healchow: Excuse me, how do you get the result of 16383 here?

xuesongxs: > Excuse me, how do you get the result of 16383 here?
> 
> If you consider the utf8mb4 character set, it takes up to 4 bytes to represent 1 character, so 65535 / 4 = 16383, then the maximum length of the char above is 255, which is also inaccurate.

What is the maximum length of the varchar field in MySQL version 5.7? Under different character sets, the disk space occupied by a character varies, so:

Under the Litan1 character set, varchar can store up to 65535-2-1=65532 characters. If it is a non empty varchar field, it can store 65535-2=65533 characters.

Under the utf8 character set, varchar can store up to floor ((65535-2-1)/3)=21844 characters. If it is a non empty varchar field, it can store floor ((65535-2)/3)=21844, and after rounding down, it is also 21844 characters

Under the utf8mb4 character set, varchar can store up to floor (65535-2-1)/4)=16383 characters. If it is a non empty varchar field, it can store 16383 characters as well.
13d831694265891853e93a00eaa2e53

healchow: So, do we need to limit the length of the CHAR type to `(255-2-1)/4=63`?

xuesongxs: > So, do we need to limit the length of the CHAR type to `(255-2-1)/4=63`?

In MySQL, char is a data type used to store fixed length strings. The maximum length of a char type is 255 characters, and smaller lengths can also be set. Among them, a fixed length string refers to a fixed length that is not affected by the stored data.

healchow: I found that in MySQL 8.0.30 on macOS, the default charset is utf8mb4, in this case, the max length of the CHAR type is 255, and the max length of the VARCHAR type is 16383, so this PR is good to me.

Charset:

image

CHAR type:

image

VARCHAR type:
image

healchow (approved):

dockerzhang (approved):

Issue #6190, opened 2022-10-14
[Improve][SDK] Update DataProxy config 


GanfengTan:
### Description

Add protocolType param

### InLong Component

InLong SDK

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #8408, opened 2023-07-04
[INLONG-7766][DataProxySDK] Adjusted frame length exceeds occurred when reporting data through the HTTP protocol

### LINKED ISSUE #7766: [Bug][SDK] Adjusted frame length exceeds occurred when reporting data through the HTTP protocol
### What happened

When using dataproxy-sdk to report data through the http protocol, the following screenshots report errors:

The scenario description is as follows (SDK only reports one piece of data per startup process):

1. This exception will be reported for the first piece of data reported after the creation of the data stream

2. If the data is not reported for a long time, the first piece of data will report this exception when the data is reported again

Guess the reason for this problem is due to the DataProxy's channel management policy configuration for netty

image

Normal reporting of messages

### How to reproduce

repair this bug

### Environment

version:1.6.0

### InLong version

1.6.0

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #7766

1. Based on the modification requirements of [a], when initializing the sending objects DefaultMessageSender and HttpProxySender, it is mandatory to add the corresponding protocol type setting call;
2. Set the protocol type explicitly in the example

a. https://github.com/apache/inlong/issues/6190

### DIFF
--- a/inlong-sdk/dataproxy-sdk/src/main/java/org/apache/inlong/sdk/dataproxy/DefaultMessageSender.java
+++ b/inlong-sdk/dataproxy-sdk/src/main/java/org/apache/inlong/sdk/dataproxy/DefaultMessageSender.java
@@ -17,6 +17,7 @@
 
 package org.apache.inlong.sdk.dataproxy;
 
+import org.apache.inlong.common.constant.ProtocolType;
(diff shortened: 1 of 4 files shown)


healchow (approved):
LGTM

PR #8698, opened 2023-08-10
[INLONG-8691][Security] Upgrade Node-Fetch Version

liaosunny123:
### Prepare a Pull Request

- Fixes #XYZ

### Motivation

Upgrade Node-Fetch Version to avoid the security problems.

### Modifications

Upgrade Node-Fetch Version

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation


leezng:
`node-fetch` is a dependency of `umi-request`, so it is a sub-dependency of this project, we should not write it directly into `package.json`, which may cause version conflicts.
I suggest that the best solution to this problem is to wait for the direct dependencies to be upgraded.

Issue #8684, opened 2023-08-09
[Improve][Security] Vite Server Options (server.fs.deny) can be bypassed using double forward-slash (//)


dockerzhang:
### Description

![image](https://github.com/apache/inlong/assets/18047329/d608babc-37cf-46ad-b4fd-579e16e15cdd)

### InLong Component

Other for not specified component

### Are you willing to submit PR?

- [ ] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

H0924cc:
I would like to work on this issue

PR #8707, opened 2023-08-11
[INLONG-8684][Dashboard] Vite Server Options (server.fs.deny) can be bypassed using double forward-slash (//)

H0924cc:
### Prepare a Pull Request
- Fixes https://github.com/apache/inlong/issues/8684

### Motivation

To solve the problem that Vite Server Options (server.fs.deny) can be bypassed using double forward-slash (//).

### Modifications

Upgrade vite to version 3.2.7 

### Verifying this change

### Documentation

  - Does this pull request introduce a new feature?  no

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,24 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
-  ~ Licensed to the Apache Software Foundation (ASF) under one or more
-  ~ contributor license agreements.  See the NOTICE file distributed with
-  ~ this work for additional information regarding copyright ownership.
-  ~ The ASF licenses this file to You under the Apache License, Version 2.0
-  ~ (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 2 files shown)


.idea/vcs.xml
```diff
@@ -1,24 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
```
healchow, line 2: Excuse me, but why change this file?

inlong-dashboard/package-lock.json
```diff
@@ -15797,7 +15797,7 @@
       "dev": true
     },
     "vite": {
-      "version": "3.0.9",
+      "version": "3.2.7",
```
leezng, line 15800: It is incomplete to modify `package-lock.json` directly, and `vite` is also a dependent package in `package.json`.

You may need to have more modifications.

leezng (changes requested):

PR #8759, opened 2023-08-17
[INLONG-8758][DataProxy] Metadata synchronization management optimization

### LINKED ISSUE #8758: [Improve][DataProxy] Metadata synchronization management optimization
### Description

In the running environment, there are the following problems that need to optimize the metadata management of DataProxy:
1. When DataProxy synchronizes metadata from Manager, it may take too long to pull data. We add a pull time consumption alarm setting to remind when encountering such an abnormal situation during use;
2. When DataProxy synchronizes metadata from the Manager and loads data from the metadata, it may be out of sync. We need to improve to avoid the situation where the metadata that is dropped to the file has not had time to load and take effect;
3. For the metadata synchronized by DataProxy, if the topic configuration is deleted but the message has already entered the channel from the source and is waiting to be sent, this part of the data should continue to be sent instead of being discarded.

### InLong Component

InLong DataProxy

### Code of Conduct

gosonzhang:
- Fixes #8758

1. Add metadata pull time consumption check and log output;
2. Increase the concurrent check of metadata loading and updating;
3. Distinguish metadata management on the source and sink sides, the source side deletes immediately, and the sink only adds but does not delete;
4. Other viewing issues modified


baomingyu (approved):
+1

healchow (approved):

PR #8792, opened 2023-08-24
[INLONG-8788][Dashboard] Data synchronization page optimization

bluewang:
### Prepare a Pull Request

- Fixes https://github.com/apache/inlong/issues/8788

### Motivation

### Modifications

SourceField and SinkField are aligned with the width of source and sink above.
Set the timeout period when the group re-executes the log operation, and the responsible person defaults to the current user when creating a new synchronization.

### Verifying this change
image

image

image


bluewang:
> Hi, could you provide more description?

ok

leezng (approved):

dockerzhang (approved):

PR #8822, opened 2023-08-30
[INLONG-7056][Manager] Support add more properties for submitting flink job

### LINKED ISSUE #7056: [Feature][Sort] Adjust sort resources according to data scale
### Description

Currently, the total amount of resources for the Flink Sort Job comes from the configuration file `flink-sort-plugin.properties`, so all submitted sort jobs will use the same amount of resources. When the data scale is large, the resources will be insufficient. When the data scale is small, the resources are wasted.

```properties
# Flink parallelism
flink.parallelism=1
```

Therefore, dynamically adjusting the number of resources according to the amount of data is one of the urgently needed functions

#### Resource Adaptive Adjustment

Theoretically, the processing performance of Flink can reach about `1000/second/core`, of course, it depends on factors such as state-backed.

Influencing factors:

- Data scale: 
- Storage IO bottleneck: 
  When the performance of a single client connection to external storage becomes a bottleneck, it is a good idea to increase the degree of parallelism or the number of threads
- Transformation computational complexity: 
  In the case of a fixed `LoadNode`, it is a deterministic factor
- Advance factors: 
  core/task manager, parallelism/core, and so on.

### Use case

_No response_

### Code of Conduct

NorthDataEngineer:
### Prepare a Pull Request

- Fixes #7056 

### Motivation

### Modifications

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-manager/manager-plugins/base/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
+++ b/inlong-manager/manager-plugins/base/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
@@ -26,6 +26,8 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileReader;
 import java.io.IOException;
+import java.lang.reflect.Field;
+import java.util.Map;
 import java.util.Properties;
 
 import static org.apache.inlong.common.constant.Constants.METRICS_AUDIT_PROXY_HOSTS_KEY;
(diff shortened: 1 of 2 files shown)


inlong-manager/manager-plugins/base/src/main/java/org/apache/inlong/manager/plugin/util/FlinkConfiguration.java
```diff
@@ -83,6 +85,30 @@ private String formatPath() throws Exception {
         return confPath;
     }
 
+    /**
+     * set flink config from Map k-v
+     */
+    public void setFlinkConfig(Map<String, String> flinkConfigMap) throws Exception {
```
healchow, line 91: Great, but how to use this method?

vernedeng:
This PR will be closed since it has been open for 1 year with no activity.

PR #8872, opened 2023-09-10
[INLONG-8871][TubeMQ] Use an error code in checkMessageAndStatus() to return the check result instead of throwing an exception

### LINKED ISSUE #8871: [Improve][TubeMQ] Use an error code in checkMessageAndStatus() to return the check result instead of throwing an exception
### Description

In the SimpleMessageProducer.checkMessageAndStatus() function [1], when checking the message parameter content, if something is abnormal, an exception will be thrown. In actual use, if the specific error can be clearly informed through the error, the API call will be more convenient to use. For example, if the message packet is too long, the caller can clearly understand the cause of the error without parsing the thrown exception information.
image

1. https://github.com/apache/inlong/blob/master/inlong-tubemq/tubemq-client/src/main/java/org/apache/inlong/tubemq/client/producer/SimpleMessageProducer.java#L300

### InLong Component

InLong TubeMQ

### Code of Conduct

gosonzhang:
- Fixes #8871 

1. Add new error code in TErrCodeConstants.java;
2. Use an error code in SimpleMessageProducer.checkMessageAndStatus() to return the check result instead of throwing an exception
3. In the SimpleMessageProducer.sendMessage() function, returns the result of the checkMessageAndStatus() check failure

### DIFF
--- a/inlong-tubemq/tubemq-client/src/main/java/org/apache/inlong/tubemq/client/producer/MessageSentResult.java
+++ b/inlong-tubemq/tubemq-client/src/main/java/org/apache/inlong/tubemq/client/producer/MessageSentResult.java
@@ -32,6 +32,13 @@ public class MessageSentResult {
     private long appendTime = TBaseConstants.META_VALUE_UNDEFINED;
     private long appendOffset = TBaseConstants.META_VALUE_UNDEFINED;
 
+    public MessageSentResult(Message message, boolean success, int errCode, String errMsg) {
+        this.message = message;
+        this.success = success;
+        this.errCode = errCode;
+        this.errMsg = errMsg;
+    }
+
     public MessageSentResult(boolean success, int errCode, String errMsg,
             Message message, long messageId, Partition partition) {
         this.success = success;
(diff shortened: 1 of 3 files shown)


aloyszhang:
This PR introduces the compatible problem because it changes the default behavior for exception processing of send message.

EMsnap:
Maybe it's better to deprecate the former method and create a new one since users may have catched the exception before

gosonzhang:
@aloyszhang, no, the calling method before and after the API remains the same, and the checked errors will not be missed. If it is sent synchronously, the checked errors will be returned in the return result; if it is sent asynchronously, the callback function will return the checked errors.

gosonzhang:
@EMsnap, if the caller only catches exceptions without checking whether the processing return result has failed, this implementation logic itself is a bug: throwing exceptions and returning error codes are only changes in the return method of the discovered problems, they both express processing errors and sending failures. The caller need captures exception and check the call return result

dockerzhang (approved):

EMsnap (approved):

PR #9116, opened 2023-10-26
[INLONG-7784][Tool] Support parallel compilation and packaging

### LINKED ISSUE #7784: [Feature][Tool] Support parallel compilation and packaging
## There are two problems with the current project:
1. When using mvn -B or mvnd to compile the project, various problems will occur and the compilation will fail.
2. The CR pipeline runs too slowly

## To locate the above problem, the analysis is as follows:

1. Reasons why parallel packaging cannot be done

The `inlong-distribution` module uses assembly to compress and package the output of each module together. However, the module dependencies are not clearly specified in the pom file, causing maven to infer dependencies.

2. CR pipeline does not need to generate distribution

## Transformation method:

- [ ] #8242 
1. CR pipeline does not package Distribution
2. Clear distribution modules and all dependencies in the project

> At present, it is being transformed to support multiple versions of Sort Flink, and it is also necessary to clarify the maven coordinates of each Connector

huanzhang:
### Prepare a Pull Request
- Fixes #7784

### Motivation

When using mvn -B or mvnd to compile the project, various problems will occur and the compilation will fail. The CR pipeline runs too slowly.

Reasons why parallel packaging cannot be done?
1. The inlong-distribution module uses assembly to compress and package the output of each module together. However, the module dependencies are not clearly specified in the pom file, causing maven to infer dependencies.
2. CR pipeline does not need to generate distribution

### Modifications
1. Remove package Distribution from CR pipeline
2. Clear distribution modules and all dependencies in the project

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -27,7 +27,6 @@ on:
       - 'inlong-common/**'
       - 'inlong-dashboard/**'
       - 'inlong-dataproxy/**'
-      - 'inlong-distribution/**'
       - 'inlong-manager/**'
       - 'inlong-sdk/**'
       - 'inlong-sort/**'
@@ -44,7 +43,6 @@ on:
       - 'inlong-common/**'
       - 'inlong-dashboard/**'
       - 'inlong-dataproxy/**'
-      - 'inlong-distribution/**'
       - 'inlong-manager/**'
       - 'inlong-sdk/**'
       - 'inlong-sort/**'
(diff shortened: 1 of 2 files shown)


.github/workflows/ci_build.yml
```diff
@@ -107,25 +105,3 @@ jobs:
         run: |
           version=`mvn -Dexec.executable='echo' -Dexec.args='${project.version}' --non-recursive exec:exec -q`
           echo "VERSION=${version}" >> $GITHUB_ENV
-
-      - name: Upload binary package
```
healchow, line 111: Excuse me, but how can we upload the binary package after removing it?

aloyszhang:
https://github.com/apache/inlong/pull/11440 support parallel building

PR #9551, opened 2024-01-08
[INLONG-9550][Dashboard] System operation and maintenance supports query and audit through Gorupid and Stream id

bluewang:
### Prepare a Pull Request

- Fixes https://github.com/apache/inlong/issues/9550

### Motivation

### Modifications

### Verifying this change

image
image

### DIFF
--- a/inlong-dashboard/src/configs/menus/conf.tsx
+++ b/inlong-dashboard/src/configs/menus/conf.tsx
@@ -101,6 +101,10 @@ const conf: MenuItemType[] = [
         path: '/system',
         name: i18n.t('configs.menus.ModuleAudit'),
       },
+      {
+        path: '/systemId',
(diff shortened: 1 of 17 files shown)


justinwwhuang (approved):

leezng:
How about using tabs instead? These two actually belong to the same module. I don’t think they need to be separated by menus.

Or in the same table, you can add a filter: dimension (IP or ID)

bluewang:
> How about using tabs instead? These two actually belong to the same module. I don’t think they need to be separated by menus.
> 
> Or in the same table, you can add a filter: dimension (IP or ID)

ok, I have made changes based on your suggestions.

leezng (approved):

dockerzhang (approved):

Issue #9884, opened 2024-03-28
[Improve] Optimized code AuditServiceImpl.java


balloon72:
### Description

What happened
Optimized code AuditServiceImpl.java

What you expected to happen
Optimized code AuditServiceImpl.java

How to reproduce
Optimized code AuditServiceImpl.java

Environment
Optimized code AuditServiceImpl.java

InLong Component
inlong-manager

### InLong Component

InLong Manager

### Are you willing to submit PR?

- [X] Yes, I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #9888, opened 2024-03-28
[INLONG-9886][Manager] Optimized code DataProxyConfigRepositoryV2

balloon72:
Title Name: [INLONG-9881] Optimized code DataProxyConfigRepositoryV2

https://github.com/apache/inlong/issues/9886

Motivation
Explain here the context, and why you're making that change. What is the problem you're trying to solve.

Modifications
Describe the modifications you've done.

PR #9896, opened 2024-03-29
[INLONG-9884][Manager] Optimized code AuditServiceImpl

balloon72:
Title Name: [INLONG-9884][Manager] Optimized code AuditServiceImpl

https://github.com/apache/inlong/issues/9884

Motivation
Explain here the context, and why you're making that change. What is the problem you're trying to solve.

Modifications
Describe the modifications you've done.


aloyszhang:
@balloon72 Thanks for your contribution!
This is a code-optimization type pull request same as #9888. Other places may also have the same problem. There will be too many issues and pull requests if we resolve them individually, also, it's hard to track all the issues.

So, I recommend bringing an umbrella issue to track all the code optimization issues at the module or sub-module level, each module or sub-module has only one issue to deal with the code optimization.
WDYT? Please feel free to share your opinions.

balloon72:
Great idea, I completely agree with this approach. Next time, I will create an umbrella issue to track code optimization problems.

aloyszhang (approved):

dockerzhang (approved):

fuweng11 (approved):

PR #10079, opened 2024-04-25
[INLONG-10066][DataProxy] Optimize the related configuration settings in the CommonConfigHolder.java file

### LINKED ISSUE #10066: [Improve][DataProxy] Optimize the related configuration settings in the CommonConfigHolder.java file
### Description

After using DataProxy for a while, I found that the definition of configuration-related keys in CommonConfigHolder.java is not very reasonable. Some names are mixed with upper and lower case, some are in camel case, and some keys belong to the same type of operation but are not categorized by name, etc. In terms of configuration settings, some configurations are too complex, making it inconvenient for users.

These issues are not user-friendly and hinder the user experience. I plan to optimize the configuration definition based on my usage experience, adjusting it according to the categories and themes of usage, such as unifying lowercase, separating key names with ".", allowing users to quickly understand which type of configuration definition it is based on the name.

However, I also noticed that some of the configuration definitions and usage in this class are not settings of the DataProxy node itself but are defined by common components. These parts should not be modified but should be adjusted at a higher level.

### InLong Component

InLong DataProxy

### Code of Conduct

The sub-issues are as follows:
- [x] #10067

gosonzhang:
Fixes #10066 

1. Uniformly configure the Key to lowercase, and use "." to classify;
2. Configure pre-reading according to different types;
3. Optimize the acquisition of configuration values, and check whether key configurations are empty and legal;

### DIFF
--- a/inlong-dataproxy/conf/common.properties
+++ b/inlong-dataproxy/conf/common.properties
@@ -18,24 +18,24 @@
 
 # manager open api address and auth key
 manager.hosts=127.0.0.1:8083
-manager.auth.secretId=
-manager.auth.secretKey=
+manager.auth.secret.id=
+manager.auth.secret.key=
 
 # proxy cluster name
 proxy.cluster.name=default_dataproxy
 proxy.cluster.tag=default_cluster
-proxy.cluster.extTag=default=true
-proxy.cluster.inCharges=admin
+proxy.cluster.ext.tag=default=true
+proxy.cluster.incharges=admin
 # synchronize interval of meta config (millisecond)
 meta.config.sync.interval.ms=10000
 # whether to startup using the local metadata.json file without connecting to the Manager
-startup.using.local.meta.file.enable=false
+meta.config.startup.using.local.file.enable=false
 
 # metric config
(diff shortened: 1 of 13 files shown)


luchunliang (approved):

inlong-dataproxy/conf/common.properties
```diff
@@ -18,24 +18,24 @@
...
 # whether to startup using the local metadata.json file without connecting to the Manager
-startup.using.local.meta.file.enable=false
+meta.config.startup.using.local.file.enable=false
 
 # metric config
 metricDomains=DataProxy
 metricDomains.DataProxy.domainListeners=org.apache.inlong.dataproxy.metrics.prometheus.PrometheusMetricListener
 metricDomains.DataProxy.snapshotInterval=60000
-prometheusHttpPort=9081
+online.metric.prometheus.http.port=9081
```
dockerzhang, line 38:
```suggestion
metric.prometheus.http.port=9081
```

dockerzhang (approved):

baomingyu (approved):

PR #10390, opened 2024-06-11
[INLONG-10261][Manager] Support installing agents by SSH key-based authentication

### LINKED ISSUE #10261: [Feature][Manager] Support installing agents by SSH key-based authentication
### Description

- SSH Authentication by username and password implemented in https://github.com/apache/inlong/pull/10098
- The Dashboard also supported the password SSH Authentication in https://github.com/apache/inlong/pull/10244

But in the product environments, most users like using key-based authentication to log in to nodes by SSH, we hope this issue can complete it.

### Use case

_No response_

### Code of Conduct

shuiqingliu:
Fixes #10261 

### Motivation
Implement SSH key-based authentication when installing the agent via SSH.

The implementation plan is as follows:
 1. On the dashboard node, add an option to distinguish between username/password and SSH key authentication when selecting the SSH method. For SSH key authentication, the user also needs to input the username, port, and IP information.
 2. When the user selects SSH key authentication, the dashboard actively calls the SSH key retrieval interface.
 3. If the manager node does not have an SSH key pair, it calls ssh-keygen to generate one. If a key pair already exists, it is returned directly.
 4.  When executing the installation command, if no user password is provided, the installation defaults to using SSH key authentication.

### Modifications
* Added an SSH key generation interface.
* Modified the remote command execution method to support SSH key authentication.
* When the user authenticates via SSH key, the password is empty, so the remote command execution method was modified to support nullable parameters.

### Verifying this change

 * Tested SSH key generation; the key was generated successfully.
image

* Tested command execution via SSH key; the execution was successful.

### Documentation

  - If yes, how is the feature documented? (not documented)

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
@@ -403,9 +403,9 @@ public interface InlongClusterService {
     /**
      * Obtain the SSH public key from the manager to install the agent.
      *
-     * @return ssh public key
+     * @return the SSH public key
      */
-    String getManagerSshPublicKey();
+    String getManagerSSHPublicKey();
 
     /**
      * Query data proxy nodes by the given inlong group id and protocol type
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
@@ -401,9 +401,15 @@ public interface InlongClusterService {
     Boolean deleteNode(Integer id, UserInfo opInfo);
 
     /**
-     * Obtain the SSH public key from the manager to install the agent.
+     * Retrieves the SSH public key from the manager node for agent installation.
      *
-     * @return the SSH public key
+     * <p>To enable SSH key-based authentication, the manager node's public key must be added to the agent node.
+     * This method either generates a new public key or retrieves an existing one.</p>
+     *
+     * <p>After obtaining the SSH public key, add it to the <code>~/.ssh/authorized_keys</code> file on the agent node.
+     * This allows the manager node to execute remote commands on the agent node via SSH.</p>
+     *
+     * @return the SSH public key as a String
      */
     String getManagerSSHPublicKey();
 
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
@@ -401,6 +401,13 @@ public interface InlongClusterService {
     Boolean deleteNode(Integer id, UserInfo opInfo);
 
     /**
+     * Obtain the SSH public key from the manager to install the agent.
+     *
+     * @return ssh public key
+     */
+    String getManagerSshPublicKey();
+
+    /**
      * Query data proxy nodes by the given inlong group id and protocol type
      *
      * @param inlongGroupId inlong group id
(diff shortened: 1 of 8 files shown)


justinwwhuang (approved):

inlong-manager/manager-web/src/main/resources/exec_cmd.exp
```diff
@@ -24,8 +24,12 @@ set cmdTimeout [lindex $argv 3]
 set runCommand [lindex $argv 4]
 set remotePort [lindex $argv 5]
 
-spawn ssh -p ${remotePort} ${remoteHost} -l ${remoteUser} "${runCommand} && echo \\#SUCCESS\\# || echo \\#fail\\#"
-
+#use public key identify
```
dockerzhang, line 27:
```suggestion
# use public key to identify
```

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cmd/shell/ShellExecutorImpl.java
```diff
@@ -92,11 +93,26 @@ private static boolean HasException(String str) {
         return false;
     }
 
+    public void syncExec(String shellPath, boolean includeEmptyParams, String... params) {
```
fuweng11, line 96: Why add this parameter, and it is always true？

shuiqingliu: Based on the current design of the `execRemote` method, it calls the expect script using hard-coded values. This script receives a password parameter. If the password parameter is empty, it will be removed by the original merge method, causing parameter mapping errors in the expect script. Currently, the SSH key authentication method does not require a password, so it is indeed necessary to retain the empty string parameter when invoke the exec_cmd.exp script.

fuweng11: If the string is allowed to be empty, then modify the deletion logic of the original merge method instead of setting this parameter.

shuiqingliu: Okay, I will remove the logic that deletes empty parameters. If we need to delete parameters in the future, we can make the corresponding changes then

inlong-manager/manager-web/src/main/java/org/apache/inlong/manager/web/controller/InlongClusterController.java
```diff
@@ -297,6 +297,13 @@ public Response<Boolean> unloadNode(@PathVariable Integer id) {
         return Response.success(clusterService.unloadNode(id, LoginUserUtils.getLoginUser().getName()));
     }
 
+    @RequestMapping(value = "/cluster/node/getManagerSSHPublicKey", method = RequestMethod.GET)
+    @ApiOperation(value = "Obtain the SSH public key from the manager to install the agent.")
+    @OperationLog(operation = OperationType.GET, operationTarget = OperationTarget.CLUSTER)
```
fuweng11, line 302: Remove it.`@OperationLog` is used to save operation logs, and this interface does not require this record.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/cluster/InlongClusterService.java
```diff
@@ -400,6 +400,13 @@ List<ClusterNodeResponse> listNodeByGroupId(
      */
     Boolean deleteNode(Integer id, UserInfo opInfo);
 
+    /**
+     * Obtain the SSH public key from the manager to install the agent.
```
dockerzhang, line 404: please add more details about what should do after getting the public key.

fuweng11 (dismissed):

dockerzhang (approved):

justinwwhuang (approved):

PR #10453, opened 2024-06-19
[INLONG-10312][Sort] Iceberg sink support report audit information exactly once

### LINKED ISSUE #10312: [Feature][Sort] Iceberg sink support report audit information exactly once
### Description

Currently, audit information reporting only supports at least once. In order to improve the defects of at least once, audit information reporting needs to be upgraded to exactly once.

image

As shown in the figure above, audit information reporting will be repeated

### Use case

Ensure that audit information report is exactly once when an operator encounters an exception or a snapshot fails.

### Code of Conduct

XiaoYou201:
### [INLONG-10312][Sort] Iceberg sink support report audit information exactly once

Fixes #10312 

### Motivation

Now，every connectors report audit information only support at least once. So，This pr base it to implements report audit information exactly once. The audit information will not be wrong in much exception situation.

### Modifications
1. Using the checkpoint principle in Flink to modify the process, the modified flow chart is shown in Figure 1 and Figure 2. In Figure 1, the callback method of notifyCompleteCheckpoint is used to upload audit information instead of scheduled upload.Each Source/Sink will save the checkpointId of the currently ongoing checkpoint, which is nowCheckpointId in the figure. When the audit information is written to the Buffer, nowCheckpointId will be attached, indicating that the audit information
is written during this ongoing checkpoint. The audit information and checkpointId are in a many-to-one relationship.

image
When a snapshot request is received, the current operator's nowCheckpointId is updated to (snapshot) checkpointId + 1. When all operators in a task complete the snapshot, the notifyCompleteCheckpoint method is called back.

At this time, AuditBuffer uploads audit information less than or equal to checkpointId (parameters in the notifyCompleteCheckpoint method).

image

2. The getCurConsumedPartitions method gets the partitions assigned to the client by the tube server, including the partitions where the client has consumed data and t the client has not consumed data. According to the previous logic, the offsets of the partitions that have not been consumed are not recorded. Here, the offsets of the partitions that have not been consumed are added.


inlong-sort/sort-flink/sort-flink-v1.15/sort-connectors/iceberg/src/main/java/org/apache/inlong/sort/iceberg/sink/IcebergStreamWriterMetrics.java
```diff
@@ -109,14 +109,26 @@ void flushDuration(long flushDurationMs) {
     }
 
     void outputMetricsWithEstimate(int size, long time) {
-        if (sourceMetricData != null) {
-            sourceMetricData.outputMetrics(1, size, time);
+        if (sinkExactlyMetric != null) {
+            sinkExactlyMetric.invokeWithId(1, size, time);
```
EMsnap, line 113: what is Id here ? can't tell it is checkpoint id by method name. just invoke is fine since the class name is exactly metric

XiaoYou201: I forgot to rename it, fixed it, thx~

vernedeng (approved):

EMsnap (approved):

PR #10511, opened 2024-06-25
[INLONG-10358][Sort] Make pulsar source support report audit information exactly once

### LINKED ISSUE #10358: [Feature][Sort] Pulsar source support report audit information exactly once
### Description

Currently, audit information reporting only supports at least once. In order to improve the defects of at least once, audit information reporting needs to be upgraded to exactly once.

image

As shown in the figure above, audit information reporting will be repeated

### Use case

Ensure that audit information report is exactly once when an operator encounters an exception or a snapshot fails.

### Code of Conduct

XiaoYou201:
### [INLONG-10358][Sort] Make pulsar source support report audit information exactly once

Fixes #10358 

### Modifications
1. Using the checkpoint principle in Flink to modify the process, the modified flow chart is shown in Figure 1 and Figure 2. In Figure 1, the callback method of notifyCompleteCheckpoint is used to upload audit information instead of scheduled upload.Each Source/Sink will save the checkpointId of the currently ongoing checkpoint, which is nowCheckpointId in the figure. When the audit information is written to the Buffer, nowCheckpointId will be attached, indicating that the audit information
is written during this ongoing checkpoint. The audit information and checkpointId are in a many-to-one relationship.

image
When a snapshot request is received, the current operator's nowCheckpointId is updated to (snapshot) checkpointId + 1. When all operators in a task complete the snapshot, the notifyCompleteCheckpoint method is called back.

At this time, AuditBuffer uploads audit information less than or equal to checkpointId (parameters in the notifyCompleteCheckpoint method).

image

2. The getCurConsumedPartitions method gets the partitions assigned to the client by the tube server, including the partitions where the client has consumed data and the client has not consumed data. According to the previous logic, the offsets of the partitions that have not been consumed are not recorded. Here, the offsets of the partitions that have not been consumed are added.

### DIFF
--- a/inlong-sort/sort-flink/sort-flink-v1.15/sort-connectors/pulsar/src/main/java/org/apache/inlong/sort/pulsar/source/PulsarSource.java
+++ b/inlong-sort/sort-flink/sort-flink-v1.15/sort-connectors/pulsar/src/main/java/org/apache/inlong/sort/pulsar/source/PulsarSource.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.inlong.sort.pulsar.source;
+
+import org.apache.flink.annotation.Internal;
+import org.apache.flink.annotation.PublicEvolving;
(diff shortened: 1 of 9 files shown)


aloyszhang:
What's the difference between `PulsarSource.java`, `PulsarSourceReaderBase`, and `PulsarSourceBuilder` and the original classed in flink-pulsar-connector?

XiaoYou201:
> What's the difference between `PulsarSource.java`, `PulsarSourceReaderBase`, and `PulsarSourceBuilder` and the original classed in flink-pulsar-connector?

There are Almost no difference between the above three classes and original classes， but I modify PulsarReader, and The PulsarTableSource create PulsarSource Through PulsarSource Builder, PulsarSource create PulsarReader. So, if I want to use custom PulsarReader, I have to modify the method which create PulsarReader in PulsarSource, and modify PulsarSourceBuilder method to build PulsarSource which can create custom PulsarReader.As for PulsarSourceBuilder, it not public, So, I have to copy it out.

aloyszhang:
I think you mean `PulsarSourceReaderBase` is not public, not `PulsarSourceReaderBase `.

aloyszhang (approved):

XiaoYou201:
> I think you mean `PulsarSourceReaderBase` is not public, not `PulsarSourceReaderBase `.

yeah~, I  correct my description , thx for your suggestion and review~~

EMsnap:
Please test on the situation where the job start from the early checkpoint with the former version of Pulsar connector

EMsnap (approved):

PR #10755, opened 2024-08-06
[INLONG-10754][Manager] Add start and ent time check for schedule info

### LINKED ISSUE #10754: [Bug][Manager] Offline data sync may create too many scheduling instances at the start point
### What happened

When configuring an offline data sync with crontab scheduling, there may be too many scheduling instances (Flink batch job), including instances before the current time.

The schedule instances before the current time should not be generated.

### How to reproduce

1. add an inlong group for offline sync with crontab expression in which the `start time` is  before the current time in the schedule info by open API
2. submit the group and approve it
3. the scheduling instances before the current time are also generated

### Environment

_No response_

### InLong version

master

### InLong Component

InLong Manager

### Code of Conduct

aloyszhang:
Fixes #10754 

### Motivation

Fix the problems for crontab expression offline sync job may generate too many schedule instances before the current time

### Modifications

1. check start time and end time, make sure not null
2. make sure the start time is not before the current time
3. make sure the end time is not before the start time

### Verifying this change

### Documentation

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
@@ -104,6 +104,7 @@ import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
 
+import java.sql.Timestamp;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


dockerzhang (approved):

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
```diff
@@ -221,6 +222,20 @@ public String save(InlongGroupRequest request, String operator) {
 
         // save schedule info for offline group
         if (DATASYNC_OFFLINE_MODE.equals(request.getInlongGroupMode())) {
+            Timestamp startTime = request.getStartTime();
+            Timestamp endTime = request.getEndTime();
+            if (startTime == null || endTime == null) {
+                LOGGER.error("start time or end time cannot be empty for groupId={}", groupId);
+                throw new BusinessException("start time or end time cannot be empty");
+            }
```
fuweng11, line 230:
```suggestion
            Preconditions.expectTrue(startTime != null && endTime != null, "start time or end time cannot be empty");
```

aloyszhang: fixed, PTAL

fuweng11 (dismissed):

aloyszhang:
I found that the update interface also needs this logic, and I will complete it soon.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
```diff
@@ -498,6 +523,7 @@ public String update(InlongGroupRequest request, String operator) {
 
         // save schedule info for offline group
         if (DATASYNC_OFFLINE_MODE.equals(request.getInlongGroupMode())) {
+            constrainStartAndEndTime(request);
```
fuweng11, line 526: checkStartAndEndTime

aloyszhang: I think constrain is better here since we do check and re-set both

fuweng11 (approved):

dockerzhang (approved):

PR #10801, opened 2024-08-16
[INLONG-10287][Agent] Update the Redis Source 

emptyOVO:
Fixes https://github.com/apache/inlong/issues/10287

## Motivation
Update the Redis source for agent

## Modifications
Modify TaskProfileDto and RedisTask to accept and parse the configuration.
Add RedisTask, RedisInstance, and RedisSource classes
Provide Unit Test

## Verifying this change
6983c4d37da7fa06df06467b6c591dd

### DIFF
--- a/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
+++ b/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
@@ -31,6 +31,8 @@ import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import redis.clients.jedis.Jedis;
+import redis.clients.jedis.ScanParams;
+import redis.clients.jedis.ScanResult;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
@@ -114,10 +116,19 @@ public class RedisSource extends AbstractSource {
     }
 
     private void startJedisSynchronize() {
-        for (String key : keys) {
-            String data = jedisExecuteCommand(redisCommand, key, fieldOrMember);
-            synchronizeData(data);
-        }
+        int cursor = 0;
+        // Each scan processes up to 100 keys
+        ScanParams scanParams = new ScanParams().count(100);
+        do {
+            ScanResult<String> scanResult = jedis.scan(cursor, scanParams);
+            for (String key : scanResult.getResult()) {
+                if (!keys.contains(key)) {
+                    String data = jedisExecuteCommand(redisCommand, key, fieldOrMember);
+                    synchronizeData(data);
+                }
+            }
+            cursor = scanResult.getCursor();
+        } while (cursor != 0);
     }
 
     private void synchronizeData(String data) {
--- a/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
+++ b/inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
@@ -23,38 +23,29 @@ import org.apache.inlong.agent.constant.TaskConstants;
 import org.apache.inlong.agent.metrics.audit.AuditUtils;
 import org.apache.inlong.agent.plugin.sources.file.AbstractSource;
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
-import com.google.gson.TypeAdapter;
-import com.google.gson.stream.JsonReader;
-import com.google.gson.stream.JsonWriter;
 import com.moilioncircle.redis.replicator.RedisReplicator;
 import com.moilioncircle.redis.replicator.Replicator;
-import com.moilioncircle.redis.replicator.cmd.CommandName;
-import com.moilioncircle.redis.replicator.cmd.impl.DefaultCommand;
-import com.moilioncircle.redis.replicator.cmd.parser.DefaultCommandParser;
 import com.moilioncircle.redis.replicator.event.PostRdbSyncEvent;
-import com.moilioncircle.redis.replicator.rdb.datatype.KeyStringValueHash;
-import com.moilioncircle.redis.replicator.rdb.datatype.KeyStringValueList;
-import com.moilioncircle.redis.replicator.rdb.datatype.KeyStringValueSet;
-import com.moilioncircle.redis.replicator.rdb.datatype.KeyStringValueString;
-import com.moilioncircle.redis.replicator.rdb.datatype.KeyStringValueZSet;
 import com.moilioncircle.redis.replicator.rdb.datatype.KeyValuePair;
-import com.moilioncircle.redis.replicator.rdb.datatype.ZSetEntry;
 import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import redis.clients.jedis.Jedis;
 
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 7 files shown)


inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
```diff
@@ -74,7 +162,18 @@ protected boolean isRunnable() {
 
     @Override
     protected void releaseSource() {
-
+        LOGGER.info("releasing redis source");
+        synchronized (this) {
```
justinwwhuang, line 166: There won't be multi-threaded calls here, no need to lock

emptyOVO: done

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
```diff
@@ -84,6 +181,278 @@ public boolean sourceFinish() {
...
+        redisReplicator.addCommandParser(CommandName.name("COPY"), defaultCommandParser);
+        redisReplicator.addCommandParser(CommandName.name("LMOVE"), defaultCommandParser);
+        redisReplicator.addCommandParser(CommandName.name("BLMOVE"), defaultCommandParser);
+        redisReplicator.addCommandParser(CommandName.name("ZDIFFSTORE"), defaultCommandParser);
+        redisReplicator.addCommandParser(CommandName.name("GEOSEARCHSTORE"), defaultCommandParser);
+        // since redis 7.0
+        redisReplicator.addCommandParser(CommandName.name("SPUBLISH"), defaultCommandParser);
+        redisReplicator.addCommandParser(CommandName.name("FUNCTION"), defaultCommandParser);
+        // add EventListener
+        redisReplicator.addEventListener((replicator, event) -> {
```
justinwwhuang, line 310: Sorry, can we only obtain data by subscribing to Redis messages? Suggest setting Redis command to obtain data

emptyOVO: Sorry, do you mean that EventListener will only obtain redis data that is added after startup ? when RedisReplicator connect to redis server it will first synchronizes the rdb file, then PostRdbSyncEvent event is triggered which means historical data of Redis is synchronized, then the redisReplicator switches to command incremental synchronization mode and captures every command that occurs in the Redis instance (such as SET, DEL, etc.) and synchronizes it

justinwwhuang: I hope we can also specify Redis commands to collect the data we want, rather than just synchronizing all Redis data.

emptyOVO: got it, maybe i didn't do that in the right way

emptyOVO: > I hope we can also specify Redis commands to collect the data we want, rather than just synchronizing all Redis data.

Is it a good idea to use Jedis on top of the redis replicator to operate and filter out the data we need ?

justinwwhuang: > > I hope we can also specify Redis commands to collect the data we want, rather than just synchronizing all Redis data.
> 
> Is it a good idea to use Jedis on top of the redis replicator to operate and filter out the data we need ?

yes

emptyOVO: ok, i will fix it

dockerzhang:
@emptyOVO please click `resolve` if you fix the comment, and you do not need to open/close the PR again and again.

emptyOVO:
> @emptyOVO please click `resolve` if you fix the comment, and you do not need to open/close the PR again and again.

got it

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
```diff
@@ -84,6 +182,96 @@ public boolean sourceFinish() {
...
+            sb.append("replId=").append(replId).append("&");
+        }
+        if (sb.charAt(sb.length() - 1) == '?' || sb.charAt(sb.length() - 1) == '&') {
+            sb.deleteCharAt(sb.length() - 1);
+        }
+        return sb.toString();
+    }
+
+    private void initReplicator() {
+        redisReplicator.addEventListener((replicator, event) -> {
```
justinwwhuang, line 220: Sorry, obtaining data through commands is recommended to be at the same level as obtaining data through subscriptions, rather than filtering through commands after subscribing.

emptyOVO: i fix it as obtaining data by Jedis through commands first, then use RedisReplicator to subscribe event and filtering data by command, is that correct ?

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
```diff
@@ -17,54 +17,169 @@
...
+            executor = Executors.newSingleThreadExecutor();
+            executor.execute(startRedisReplicator());
+        } catch (URISyntaxException | IOException e) {
+            sourceMetric.pluginReadFailCount.addAndGet(1);
+            LOGGER.error("Connect to redis {}:{} failed.", hostName, port);
+        }
+    }
 
+    private void startJedisSynchronize() {
+        for (String key : keys) {
```
justinwwhuang, line 117: Can data acquisition be done on a scheduled basis instead of just once.

inlong-agent/agent-plugins/src/main/java/org/apache/inlong/agent/plugin/sources/RedisSource.java
justinwwhuang: > i fix it as obtaining data by Jedis through commands first, then use RedisReplicator to subscribe event and filtering data by command, is that correct ?

1. We don't necessarily mean to collect all the data from the entire Redis cluster. Your implementation here has always wanted to collect all the data? From the earliest subscription to scanning all keys now.
2. The reids command is not used as a filter, it is a different collection method from subscription. Remember, there should be no filtering logic.

emptyOVO: > > i fix it as obtaining data by Jedis through commands first, then use RedisReplicator to subscribe event and filtering data by command, is that correct ?
> 
> 1. We don't necessarily mean to collect all the data from the entire Redis cluster. Your implementation here has always wanted to collect all the data? From the earliest subscription to scanning all keys now.
> 2. The reids command is not used as a filter, it is a different collection method from subscription. Remember, there should be no filtering logic.

then I feel like we don't need jedis at all, we can just get the data we need from subscribed events in redis replicator, Debezium does this for data capture and synchronization of databases too

justinwwhuang: > > > i fix it as obtaining data by Jedis through commands first, then use RedisReplicator to subscribe event and filtering data by command, is that correct ?
> > 
> > 
> > 
> > 1. We don't necessarily mean to collect all the data from the entire Redis cluster. Your implementation here has always wanted to collect all the data? From the earliest subscription to scanning all keys now.
> > 2. The reids command is not used as a filter, it is a different collection method from subscription. Remember, there should be no filtering logic.
> 
> then I feel like we don't need jedis at all, we can just get the data we need from subscribed events in redis replicator, Debezium does this for data capture and synchronization of databases too

No, I hope to support data retrieval without subscription, only through Redis commands.

PR #10817, opened 2024-08-20
[INLONG-10809][SDK] Improvements to TypeConverter field types and CompareValue in OperatorTools

### LINKED ISSUE #10809: [Improve][SDK] Improvements to TypeConverter field types and CompareValue in OperatorTools
### Description
- parent issue #10796 

**Data:**
```
numeric1|numeric2|numeric3|numeric4
3.14159265358979323846|4|6|8
```
**SQL Statement:**
```sql
select if(numeric2 = 4,1,0) from source;
```
**Background:**
When parsing `'numeric2=4'`, the following function will be entered:
```java
org.apache.inlong.sdk.transform.process.operator.OperatorTools#compareValue()
public static int compareValue(Comparable left, Comparable right) {
	if (left == null) {
		return right == null ? 0 : -1;
	}
	if (right == null) {
		return 1;
	}
	/*
		When one of the left and right is a string, the other is assumed to be BigDecimal, but there are other types present during the parsing process. 
		For example, here JSQLParser parses the 4 in the expression as a Long type, while numeric2 is parsed by the SDK as a string "4". In "ObjectUtils.compare()", an error occurs due to inconsistent types on both sides.
	*/
	if (left instanceof String) {
		if (right instanceof String) {
			return ObjectUtils.compare(left, right);
		} else {
			BigDecimal leftValue = parseBigDecimal(left);
			return ObjectUtils.compare(leftValue, right);
		}
	} else {
		if (right instanceof String) {
			BigDecimal rightValue = parseBigDecimal(right);
			return ObjectUtils.compare(left, rightValue);
		} else {
			return ObjectUtils.compare(left, right);
		}
	}
}
```
**Questions:**
1. The Transform reserved a field type interface, but it was not used in subsequent development.
2. The compareValue method in OperatorTools did not take into account in detail the types involved in comparing the two instances.

### InLong Component

InLong SDK

### Code of Conduct

Zkplo:
Fixes #10809 

### Motivation

- Modified the SourceData interface to enable it to return multiple data types; Based on this, the CsvSourceData class was modified.

  - The ValueParser interface has reserved the ability to return multiple data types, so SourceData only needs to be aligned with it.

- To enable the functionality of field types, modify the FieldInfo class to have a default TypeConverter.

  - If the original FieldInfo class needs to pass TypeConverter through the constructor, otherwise it will be null, causing each Field to have no type; This requires specifying the default implementation of TypeConverter.

- Based on the TypeConverter interface, two converters, DoubleConverter and LongConverter, have been implemented.

- In order to enable InLong Transform to parse floating-point data types in SQL statements, the DoubleParser class is added.

  For example, the following SQL statement:

  ```sql
  select if(numeric1 < 4.3,1,0) from source
  ```

  InLong Transform cannot parse `4.3`, resulting in the expression being parsed as `null` on the right-hand side.

- Modify the OperatorTool.compareValue method to align the types of both parties being compared. The specific processing is as follows:

  -  If the types on both sides are the same, compare them directly.
  -  If the types on both sides are different, try constructing them as the `BigDecimal` type:
     -  If the construction process passes normally, it means that numerical comparison can be made.
     -  If an exception is thrown during the construction process, indicating the existence of a non numeric type, then it should be uniformly converted to the `String` type for comparison.

- Add IfFunction class for testing various classes of ExpressionOperator.

### Verifying this change

  *(please describe tests)*

  *(example:)*

### DIFF
--- a/inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/decode/CsvSourceData.java
+++ b/inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/decode/CsvSourceData.java
@@ -28,14 +28,14 @@ import java.util.Map;
  */
 public class CsvSourceData implements SourceData {
 
-    private List<Map<String, String>> rows = new ArrayList<>();
+    private List<Map<String, Object>> rows = new ArrayList<>();
 
-    private Map<String, String> currentRow;
+    private Map<String, Object> currentRow;
 
     public CsvSourceData() {
     }
 
-    public void putField(String fieldName, String fieldValue) {
+    public void putField(String fieldName, Object fieldValue) {
         this.currentRow.put(fieldName, fieldValue);
     }
 
@@ -50,11 +50,11 @@ public class CsvSourceData implements SourceData {
     }
 
     @Override
-    public String getField(int rowNum, String fieldName) {
+    public Object getField(int rowNum, String fieldName) {
         if (rowNum >= this.rows.size()) {
             return null;
         }
-        Map<String, String> targetRow = this.rows.get(rowNum);
+        Map<String, Object> targetRow = this.rows.get(rowNum);
         return targetRow.get(fieldName);
     }
 }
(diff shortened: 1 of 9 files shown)


vernedeng:
please add more details about this issue

Zkplo:
More information has been added. @vernedeng

aloyszhang (approved):

aloyszhang (dismissed):

aloyszhang (approved):

luchunliang (approved):

PR #10904, opened 2024-08-26
[INLONG-10902][SDK] Transform support HEX(numeric or string) function

### LINKED ISSUE #10902: [Feature][SDK] Transform support HEX(numeric or string) function
### Description
parent issue https://github.com/apache/inlong/issues/10796
to support flinkSQL common function HEX(numeric or string)
* If the input argument is a numeric value (such as an integer), the HEX function converts the value to the corresponding hexadecimal string.
* If the input argument is a string, the HEX function converts each character in the string to its corresponding hexadecimal ASCII encoding and returns the hexadecimal representation of the entire string.

For example, HEX('abc') returns '616263' because 'a' is ASCII 61, 'b' is 62, and 'c' is 63.
For example, HEX(255) returns 'FF'.

### Use case

_No response_

### Code of Conduct

emptyOVO:
Fixes #10902 

### Modifications
Add a function class: HexFunction and its unit test.
to support flinkSQL common functions HEX(numeric or string)

* If the input argument is a numeric value (such as an integer), the HEX function converts the value to the corresponding hexadecimal string.
* If the input argument is a string, the HEX function converts each character in the string to its corresponding hexadecimal ASCII encoding and returns the hexadecimal representation of the entire string.
For example, HEX('abc') returns '616263' because 'a' is ASCII 61, 'b' is 62, and 'c' is 63.
For example, HEX(255) returns 'FF'.

### Verifying this change

- [X] This change is already covered by existing tests, such as:
  *(please describe tests)*

  *(example:)*

### Documentation


inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/function/HexFunction.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public HexFunction(Function expr) {
+        valueParser = OperatorTools.buildParser(expr.getParameters().getExpressions().get(0));
+    }
+
+    @Override
+    public Object parse(SourceData sourceData, int rowIndex, Context context) {
+        Object valueObj = valueParser.parse(sourceData, rowIndex, context);
+        try {
+            return Integer.toHexString(OperatorTools.parseBigDecimal(valueObj).intValue()).toUpperCase();
```
aloyszhang, line 44: Why only support integer?

emptyOVO: > Why only support integer?

The HEX() function in flinksql doesn't handle float or double directly, it cut out the integer part of it, but I think I can do a numeric check such as Long, i will fix it, do i need to add the float or double handler function? it may still cause loss of data accuracy, Some data can be very long when converted into hexadecimal and there is no suitable java type can store it

aloyszhang (dismissed):

inlong-sdk/transform-sdk/src/test/java/org/apache/inlong/sdk/transform/process/TestTransformArithmeticFunctionsProcessor.java
```diff
@@ -619,4 +619,33 @@ public void testRandFunction() throws Exception {
...
+        Assert.assertEquals(output1.get(0), "result=3EF");
+        // case: hex(3.14)
+        List<String> output2 = processor1.transform("3.14|4|6|8", new HashMap<>());
+        Assert.assertEquals(1, output2.size());
+        Assert.assertEquals(output2.get(0), "result=3");
+        // case: hex(3.141592653589793)
+        List<String> output3 = processor1.transform("3.141592653589793|4|6|8", new HashMap<>());
+        Assert.assertEquals(1, output3.size());
+        Assert.assertEquals(output3.get(0), "result=3");
+        // case: hex(9223372036854775807)
```
yfsn666, line 641:
```suggestion
        // case: hex(-9223372036854775807)
```

inlong-sdk/transform-sdk/src/test/java/org/apache/inlong/sdk/transform/process/TestTransformArithmeticFunctionsProcessor.java
```diff
@@ -619,4 +619,33 @@ public void testRandFunction() throws Exception {
...
+        List<String> output4 = processor1.transform("-9223372036854775808|4|6|8", new HashMap<>());
+        Assert.assertEquals(1, output4.size());
+        Assert.assertEquals(output4.get(0), "result=8000000000000000");
```
yfsn666, line 644: The result is not correct?HEX(-9223372036854775808) should be -8000000000000000.

inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/function/HexFunction.java
```diff
@@ -0,0 +1,93 @@
...
+        valueParser = OperatorTools.buildParser(expr.getParameters().getExpressions().get(0));
+    }
+
+    @Override
+    public Object parse(SourceData sourceData, int rowIndex, Context context) {
+        Object valueObj = valueParser.parse(sourceData, rowIndex, context);
+        try {
+            return hex(OperatorTools.parseBigDecimal(valueObj)).toUpperCase();
+        } catch (NumberFormatException e) {
+            return hex(OperatorTools.parseString(valueObj));
```
vernedeng, line 50: use `instanceof` to derive the specified method call, instead of the exception

emptyOVO: > use `instanceof` to derive the specified method call, instead of the exception
I have tried that before, but if we use `instanceof` to derive the specified method call like this
```
if (valueObj instanceof BigDecimal) {
            return hex(OperatorTools.parseBigDecimal(valueObj)).toUpperCase();
        } else if(valueObj instanceof String) {
            return hex(OperatorTools.parseString(valueObj)).toUpperCase();
        }
        return null;
```
because `valueParser.parse(sourceData, rowIndex, context);` always return a String so `if (valueObj instanceof BigDecimal)` will never reached, and i checked the `parseBigDecimal` method, it seens that it also doesn't handle the NumberFormatException, i think i should handle it here or the `parseBigDecimal` method

inlong-sdk/transform-sdk/src/test/java/org/apache/inlong/sdk/transform/process/TestTransformArithmeticFunctionsProcessor.java
emptyOVO: > The result is not correct?HEX(-9223372036854775808) should be -8000000000000000.

Because the HEX function is converted based on the binary complement representation of the numeric value.
In binary complement notation, the binary representation of -9223372036854775808 is the same as the unsigned 9223372036854775808. So, when you convert this binary number to a hexadecimal string, the result is 8000000000000000 without the negative sign
This behavior is standard in systems that use binary complement representations, including programming environments such as Java and SQL

inlong-sdk/transform-sdk/src/test/java/org/apache/inlong/sdk/transform/process/TestTransformArithmeticFunctionsProcessor.java
emptyOVO: done

inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/function/HexFunction.java
vernedeng: Maybe you can make a new method called `isBigDecimal` to check the type of input valueObj, if true, parse it in BigDecimal way; otherwise, parse it as String.

```
if (isBigDecimal(valueObj)) {
     return hex(OperatorTools.parseBigDecimal(valueObj)).toUpperCase();
} 
return hex(OperatorTools.parseString(valueObj)).toUpperCase();
```

emptyOVO: > Maybe you can make a new method called `isBigDecimal` to check the type of input valueObj, if true, parse it in BigDecimal way; otherwise, parse it as String.
> 
> ```
> if (isBigDecimal(valueObj)) {
>      return hex(OperatorTools.parseBigDecimal(valueObj)).toUpperCase();
> } 
> return hex(OperatorTools.parseString(valueObj)).toUpperCase();
> ```

i fix it as using regex to check the data

emptyOVO: done

aloyszhang (approved):

dockerzhang (approved):

vernedeng (approved):

inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/function/HexFunction.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * HexFunction
+ * description:
+ * - If the input argument is a numeric value (such as an integer), the HEX function converts the value to the corresponding hexadecimal string.
+ * - If the input argument is a string, the HEX function converts each character in the string to its corresponding hexadecimal ASCII encoding and returns the hexadecimal representation of the entire string.
+ */
+@TransformFunction(names = {"hex"})
+class HexFunction implements ValueParser {
```
luchunliang, line 39: Please set the Java keyword 'public'.

PR #11279, opened 2024-10-04
[INLONG-11208][SDK] Transform support EXISTS() operator

### LINKED ISSUE #11208: [Feature][SDK] Transform support EXISTS() operator
### Description

* parent issue: #10796
* from flinksql: https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/functions/systemfunctions/#collection-functions
* EXISTS (sub-query)--Returns TRUE if sub-query returns at least one row.

### Use case

_No response_

### Code of Conduct

emptyOVO:
Fixes #11208 

### Motivation
add ExistsParser.class ,provide unit tests

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/parser/ExistsParser.java
+++ b/inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/parser/ExistsParser.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


inlong-sdk/transform-sdk/src/main/java/org/apache/inlong/sdk/transform/process/parser/ExistsParser.java
```diff
@@ -0,0 +1,50 @@
...
+    private final ValueParser subQueryValueParser;
+
+    public ExistsParser(ExistsExpression expr) {
+        Expression rightExpression = expr.getRightExpression();
+        this.subQueryValueParser = OperatorTools.buildParser(rightExpression);
+    }
+
+    @Override
+    public Object parse(SourceData sourceData, int rowIndex, Context context) {
+        try {
```
luchunliang, line 43: [NOT] EXISTS(subquery)
The semantic scenario of using EXISTS in subqueries, where the example is similar to the semantics of IS NULL.

emptyOVO: you're right, the current implementation is incorrect, Should the parsing of subqueries be at the function/parser/operator level or at the TransformProcessor level？it seems that the `initTransformSql` method in `TransformProcessor`  doesn't 
 handle the subquery

emptyOVO:
The semantic scenario of using EXISTS in subqueries, the example is similar to the semantics of IS NULL. so this pr will be close

PR #11395, opened 2024-10-23
[INLONG-11386][TubeMQ] Use local files to save consumer group offset information

### LINKED ISSUE #11386: [Improve][TubeMQ] Use local files to save consumer group offset information
### Description

When using TubeMQ currently, we need to configure ZooKeeper to persist the offset information of consumer groups. However, it is not reasonable to use ZooKeeper to persist the offset of consumer groups. The cost is too high. The offset of consumer groups can be stored in local files.

If we can support local storage of consumer group offsets on the current basis, it will greatly reduce the cost of using our system. This issue is improved based on this idea.

### InLong Component

InLong TubeMQ

### Code of Conduct

gosonzhang:
Fixes #11386

image

Optimize content:
1. Broker regularly refreshes the consumer group offset information in memory to the file Offset Cache according to the configured offsetStgCacheFlushMs value;
2. The consumer group offset information stored in the file Offset Cache is regularly written to the offsetStgFilePath/offsetDir directory according to the configured offsetStgFileSyncMs value. The file name successfully saved is offsets.meta. The persistence logic is as follows:
  - Persistently save the consumer group offset information to the file offsets.tmp. If the write fails, output the failure log, alarm and exit the persistence process;
  - Read the offsets.tmp file and compare it with the previously written information to see if it is consistent. If it is inconsistent, output the failure log and exit the persistence process;
  - Migrate the file from offsets.tmp to offsets.mid. If the offsets.meta file currently exists, change offsets.meta to offsets.old;
  - Migrate the file from offsets.mid to offsets.meta, and delete the offsets.old file at the same time.

3. Broker supports saving the current consumer group's Offset collection information to the backupPath path specified in the API interface through the HTTP API admin_backup_group_offsets. The saved file name is offsets.meta.

### DIFF
--- a/inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corebase/TBaseConstants.java
+++ b/inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corebase/TBaseConstants.java
@@ -84,7 +84,6 @@ public class TBaseConstants {
     public static final long CFG_MIN_META_FORCE_UPDATE_PERIOD = 1 * 60 * 1000;
     public static final long CFG_STATS_MIN_SNAPSHOT_PERIOD_MS = 2000;
 
-    public static final int OFFSET_HISTORY_RECORD_VERSION = 1;
     public static final int OFFSET_HISTORY_RECORD_SHORT_VERSION = 2;
     public static final int OFFSET_HISTORY_RECORD_VERSION_3 = 3;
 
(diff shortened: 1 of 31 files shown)


inlong-tubemq/tubemq-core/src/main/java/org/apache/inlong/tubemq/corebase/TBaseConstants.java
```diff
@@ -84,4 +84,9 @@ public class TBaseConstants {
     public static final long CFG_MIN_META_FORCE_UPDATE_PERIOD = 1 * 60 * 1000;
     public static final long CFG_STATS_MIN_SNAPSHOT_PERIOD_MS = 2000;
 
+    public static final int OFFSET_HISTORY_RECORD_VERSION = 1;
```
aloyszhang, line 87: Not used.

aloyszhang (approved):

luchunliang (approved):

PR #11582, opened 2024-12-05
[INLONG-11583][TubeMQ] Go SDK load balance logic not perfect, causing consumption to stop suddenly

### LINKED ISSUE #11583: [Bug][TubeMQ] Go SDK load balance logic not perfect, causing consumption to stop suddenly
### What happened

consumption suddenly stop
image

1. If the old code fails to register to Master again, there will be no further actions. It needs to be changed to continue retrying indefinitely upon failure.
企业微信截图_17334052584263

2. OfferEventAndNotify is responsible for receiving rebalance events, while OfferEventResult handles the results of those events. Therefore, the former should definitely not operate on rebalanceResults, nor should it reprocess the header of rebalanceResults as if it were a newly received event. Otherwise, if the result of the previous event has not been reported yet when the second event arrives, the subsequent processRebalanceEvent will end up processing the first event again, causing the second received event to be discarded.
企业微信截图_1733405369322

3. ip_selector directly returns the domain name, while the connection pool maintains one connection per address, and it will not reconnect as long as the packet sending and receiving are normal. If there is a cluster switch, as long as the old master is sending and receiving packets normally, even if the registration fails, the TCP connection will not reconnect. Therefore, it is necessary to resolve the domain name into an IP address before returning it. During a cluster switch, a new IP can be returned by re-selecting. 
企业微信截图_17334056241261

4. Here, there is no check to see if rsp's success is true. If a cluster change causes the master to no longer recognize the client, it will continue to send heartbeats without re-registering.
企业微信截图_17334061728723

### How to reproduce

when master changes

### Environment

_No response_

### InLong version

master

### InLong Component

InLong TubeMQ

### Code of Conduct

yuexihan:
Fixes #11583

### Verifying this change

  *(please describe tests)*

  *(example:)*

### Documentation

### DIFF
--- a/inlong-tubemq/tubemq-client-twins/tubemq-client-go/client/consumer_impl.go
+++ b/inlong-tubemq/tubemq-client-twins/tubemq-client-go/client/consumer_impl.go
@@ -132,7 +132,7 @@ func (c *consumer) routine(ctx context.Context) {
 			return
 		default:
 		}
-		// 选主
+		// select master
 		node, err := c.selector.Select(c.config.Consumer.Masters)
 		if err != nil {
 			log.Errorf("[CONSUMER] select error %v", err)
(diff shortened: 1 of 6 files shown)


aloyszhang:
please create an issue to describe what this pull request is for.

inlong-tubemq/tubemq-client-twins/tubemq-client-go/client/consumer_impl.go
```diff
@@ -105,72 +106,92 @@ func NewConsumer(config *config.Config) (Consumer, error) {
...
+func (c *consumer) routine(ctx context.Context) {
+	defer close(c.routineClosed)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		default:
+		}
+		// 选主
```
gosonzhang, line 135: To make it easier for more people to understand, please use English annotations

gosonzhang (approved):

justinwwhuang (approved):

PR #11872, opened 2025-05-21
[INLONG-11871][SDK] Fix init crashes when no IP available

### LINKED ISSUE #11871: [Bug] Go programs importing the DataProxy SDK may crash with no way to handle the error
### What happened

https://github.com/apache/inlong/blob/master/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/util/id.go#L40

Using `log.Fatal()` in `init()` prevents programs importing the SDK to handle an eventual error. Any initialisation routine that may encounter errors should be called explicitly, and errors should be returned.

My problem: I cannot run Telegraf (https://github.com/influxdata/telegraf) on a machine with no network adapter. I do not expect InLong to work under these circumstances, but I do not expect that importing the SDK would prevent Telegraf from starting.

I expect the SDK to fail to initialise in a way that lets my program handle it. Whether the program should stop immediately or maybe continue with the InLong related features disabled should be left to my program.

### How to reproduce

Run any Go program importing the DataProxy SDK on a machine without  a network adapter.

```
import (
	"github.com/apache/inlong/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/dataproxy"
)
```

You can repro in a Docker container.
```
docker run --network none ...
```

Sample output:
```
2025/05/20 17:01:56 no ip
```

### Environment

Tested on Linux, Ubuntu 24.04 (host and container). Also encountered on Mac CI machines.

### InLong version

master

### InLong Component

InLong DataProxy

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

marcv81:
If an IP is not available, we use id = 0 instead of crashing.

Fixes #11871

### Motivation

This PR allows programs which import the InLong DataProxy SDK to initialise properly even if no IP address is available on the system. Modular programs such as Telegraf may import the InLong SDK, but enable/disable its features in the configuration. Currently, such programs crash even if the configuration disables InLong. This PR fixes that.

### Modifications

Fairly trivial, please see the diff.

### Verifying this change

- [X] This change is already covered by existing tests, such as:
  - happy path: all tests, because the change only impacts the init() function.
  - sad path: I manually verified this PR allows Telegraf to start in a Docker container with no network.

### Documentation

  - Does this pull request introduce a new feature? No

### DIFF
--- a/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/util/id.go
+++ b/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/util/id.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"log"
+	"sync"
 
 	"github.com/bwmarrin/snowflake"
 	"github.com/google/uuid"
@@ -25,20 +26,22 @@ import (
 )
 
 var (
+	snowflakeOnce sync.Once
 	snowflakeNode *snowflake.Node
+	snowflakeErr  error
 )
 
-func init() {
+func newSnowFlakeNode() (*snowflake.Node, error) {
 	ip, err := GetOneIP()
 	if err != nil {
-		log.Fatal(err)
+		return nil, err
 	}
-
 	id := IPtoUInt(ip)
-	snowflakeNode, err = snowflake.NewNode(int64(id % 1024))
+	node, err := snowflake.NewNode(int64(id % 1024))
 	if err != nil {
-		log.Fatal(err)
+		return nil, err
 	}
+	return node, nil
 }
 
 // UInt64UUID generates an uint64 UUID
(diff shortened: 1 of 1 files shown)


zak-pawel (approved):

inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/util/id.go
```diff
@@ -29,12 +29,14 @@ var (
 )
 
 func init() {
+	var id uint32 = 0
+
+	// Generate the ID from an IP if possible.
 	ip, err := GetOneIP()
-	if err != nil {
-		log.Fatal(err)
```
gunli, line 34: Since we do not use SnowFlakeID as the batchID any more, I think do not panic here is acceptable, but I think it is better to let the user know why it fails, may be we can update it like this:
```
func init() {
	ip, err := GetOneIP()
	if err != nil {
		log.Println("get local ip failed, stop to init snowflake")
                return
	}

	id := IPtoUInt(ip)
	snowflakeNode, err = snowflake.NewNode(int64(id % 1024))
	if err != nil {
		log.Println("init snowflake failed, you can not use snowflake id generator further")
	}
}

func SnowFlakeID() (string, error) {
	if snowflakeNode == nil {
		return "", errors.New("snowflake is not initialized")
	}
	return snowflakeNode.Generate().String(), nil
}
```

gunli (changes requested):

gunli (changes requested):

inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/util/id.go
marcv81: Note: I used `Lock` instead of `Once`. This way `initSnowflake` has a chance to succeed if it previously failed, instead of permanently remaining in an error state. Maybe this is helpful if the program using the SDK started while the network was down? Or maybe this is overkill :) Please let me know what you think.

gunli: Hi @marcv81 , thanks for your update. I think this is overkill, in big data area, we may send/report a lot of data in 1s, if we use SnowFlakeID as the batchID, it is a frequently call, `Lock` is inefficient. Besides, let it failed at the beginning and succeed later is not good for data consistency, it is better to fail fast at the beginning.

gunli (approved):
LGTM

aloyszhang (approved):

dockerzhang (approved):

lvanneo (approved):
LGTM

PR #11933, opened 2025-07-10
[INLONG-11929][SDK] Dataproxy GO SDK supports HTTP reporting

dark-Qy:
Fixes [#11929](https://github.com/apache/inlong/issues/11929)

### Motivation

As described in issue #11929, the InLong DataProxy Go SDK currently only supports TCP for data reporting. While the DataProxy server provides a public HTTP API for this purpose, it is not yet integrated into the Go SDK. This requires users to write their own HTTP client logic, which is inconvenient and leads to inconsistent implementations.

This pull request addresses this gap by integrating a native HTTP reporting mechanism into the Go SDK, providing users with a more flexible choice of protocols. This is particularly useful in environments where TCP traffic is restricted or when HTTP is preferred for its simplicity and firewall-friendliness.

### Modifications

This pull request introduces HTTP reporting capabilities with the following key changes:

1.  **Added `Protocol` Enum and `WithProtocol` Option**:
    -   A new `Protocol` enum (`ProtocolTCP`, `ProtocolHTTP`) has been added to allow protocol selection.
    -   The `WithProtocol(protocol Protocol)` option enables users to easily switch to HTTP during client initialization. The default remains TCP to ensure backward compatibility.

2.  **Implemented a `HTTPClient`**:
    -   A new `http_client.go` file encapsulates the `HTTPClient`, which manages a pooled `http.Client`.
    -   It handles the construction of HTTP requests according to the [DataProxy HTTP API specification](https://inlong.apache.org/docs/sdk/dataproxy-sdk/http/) and processes server responses.
    -   To ensure balanced load distribution across DataProxy nodes, it uses an atomic counter for round-robin endpoint selection.

3.  **Implemented a `HTTPWorker`**:
    -   A new `http_worker.go` file contains the `HTTPWorker`, which batches messages and sends them asynchronously.
    -   To maximize performance and resilience, the worker is fully non-blocking. When a send operation fails, it triggers an **asynchronous exponential backoff** retry mechanism, allowing it to continue processing new messages without delay.

4.  **Seamless Integration into the Main `Client`**:
    -   The main `client.go` has been updated to act as a dispatcher. Key methods like `initAll`, `Send`, `SendAsync`, and `Close` now delegate to the appropriate TCP or HTTP implementation based on the configured protocol.

### Verifying this change

  - Added `ExampleClient_Send_http` and `ExampleClient_SendAsync_http` in `dataproxy/example_test.go`. These examples demonstrate how to initialize the client with the HTTP protocol and use both synchronous (`Send`) and asynchronous (`SendAsync`) methods.
  - These examples can be executed via `go test`. They serve as both a verification of the implementation and as live documentation for users.

### Documentation

- Does this pull request introduce a new feature? **yes**
- If yes, how is the feature documented? **docs**
- This PR introduces a new feature that will be documented on the official InLong website. A follow-up issue will be created to track this documentation update. The new options (`WithProtocol`, `WithHTTPTimeout`, `WithHTTPMaxConns`) and a usage example for the HTTP protocol will be added.

### DIFF
--- a/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/dataproxy/client.go
+++ b/inlong-sdk/dataproxy-sdk-twins/dataproxy-sdk-golang/dataproxy/client.go
@@ -59,6 +59,8 @@ type client struct {
 	connPool                 connpool.EndpointRestrictedConnPool // connection pool
 	netClient                *gnet.Client                        // client side network manager
 	workers                  []*worker                           // worker to do send and receive jobs
+	httpClient               *HTTPClient                         // HTTP client instance for managing HTTP-based communication
+	httpWorkers              []*HTTPWorker                       // Slice of HTTPWorker instances for handling HTTP tasks
 	curWorkerIndex           atomic.Uint64                       // current worker index
 	log                      logger.Logger                       // debug logger
 	metrics                  *metrics                            // metrics
@@ -101,7 +103,17 @@ func (c *client) initAll() error {
 	if err != nil {
 		return err
 	}
-	err = c.initNetClient()
+
+	switch c.options.Protocol {
+	case ProtocolHTTP:
+		return c.initHTTP()
+	default:
+		return c.initTCP()
+	}
+}
+
+func (c *client) initTCP() error {
+	err := c.initNetClient()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


fuweng11:
Please provide relevant information on the test results.

PR #12107, opened 2026-03-30
[INLONG-12106][Manager] Add relevant MQ cluster checks

### LINKED ISSUE #12106: [Bug][Manager] Problems arising from the MQ cluster during deletion process
### What happened

When creating an access in the data access module, you can select the corresponding MQ type to create a data flow group.
Image
Even without a corresponding MQ cluster, new connections can still be established and submitted for approval. However, issues may arise when deleting because there is no corresponding MQ cluster.
Image
Furthermore, I believe that if there is no corresponding MQ cluster, the creation should not be successful.

Deletion process
InlongGroupController.delete(...)
-> InlongGroupProcessService.deleteProcess(String groupId, String operator)
-> InlongGroupProcessService.invokeDeleteProcess(...)
-> workflowService.start(ProcessName.DELETE_GROUP_PROCESS, operator, form)
The workflow will call DeleteGroupWorkflowDefinition.defineProcess(), which includes the DeleteMQ task in the deletion process.
This task will execute QueueResourceListener.listen(...)
In the DELETE case, it calls queueOperator.deleteQueueForGroup(groupInfo, operator);
This method has different implementations for different MQs.
In Kafka
```java
ClusterInfo clusterInfo = clusterService.getOne(groupInfo.getInlongClusterTag(), null, ClusterType.KAFKA);

@Override  
public ClusterInfo getOne(String clusterTag, String name, String type) {  
    List<InlongClusterEntity> entityList = clusterMapper.selectByKey(clusterTag, name, type);  
    if (CollectionUtils.isEmpty(entityList)) {  
        throw new BusinessException(String.format("cluster not found by tag=%s, name=%s, type=%s",  
                clusterTag, name, type));  
    }  
  
    InlongClusterEntity entity = entityList.get(0);  
    InlongClusterOperator instance = clusterOperatorFactory.getInstance(entity.getType());  
    ClusterInfo result = instance.getFromEntity(entity);  
    LOGGER.debug("success to get inlong cluster by tag={}, name={}, type={}", clusterTag, name, type);  
    return result;  
}
```
An exception will be thrown if no cluster of the corresponding type is found.
In the log
```
[ ] 2026-03-30 05:53:58.572 -ERROR [http-nio-8083-exec-10] a.i.m.w.e.LogableEventListener:88 - execute listener WorkflowEventLogEntity(id=null, processId=48, processName=DELETE_GROUP_PROCESS, processDisplayName=Delete Group, inlongGroupId=test_group_id, taskId=130, elementName=DeleteMQ, elementDisplayName=Group-DeleteMQ, eventType=TaskEvent, event=COMPLETE, listener=QueueResourceListener, startTime=Mon Mar 30 05:53:58 UTC 2026, endTime=null, status=-1, async=0, ip=172.18.0.7, remark=null, exception=cluster not found by tag=null, name=null, type=KAFKA) error:
org.apache.inlong.manager.common.exceptions.BusinessException: cluster not found by tag=null, name=null, type=KAFKA
at org.apache.inlong.manager.service.cluster.InlongClusterServiceImpl.getOne(InlongClusterServiceImpl.java:545) ~[manager-service-2.3.0.jar:2.3.0]
```
In Pulsar
```java
List<ClusterInfo> clusterInfos = clusterService.listByTagAndType(clusterTag, ClusterType.PULSAR);

@Override  
public List<ClusterInfo> listByTagAndType(String clusterTag, String clusterType) {  
    List<InlongClusterEntity> clusterEntities = clusterMapper.selectByKey(clusterTag, null, clusterType);  
    if (CollectionUtils.isEmpty(clusterEntities)) {  
        throw new BusinessException(String.format("cannot find any cluster by tag %s and type %s",  
                clusterTag, clusterType));  
    }  
  
    List<ClusterInfo> clusterInfos = clusterEntities.stream()  
            .map(entity -> {  
                InlongClusterOperator operator = clusterOperatorFactory.getInstance(entity.getType());  
                return operator.getFromEntity(entity);  
            })  
            .collect(Collectors.toList());  
  
    LOGGER.debug("success to list inlong cluster by tag={}", clusterTag);  
    return clusterInfos;  
}
```
The TubeMQ deletion process has not been implemented yet, so no exception will be triggered.

### How to reproduce

1. Create a new connection and use an MQ without a corresponding MQ cluster.
2. Delete it

### Environment

_No response_

### InLong version

2.3.0

### InLong Component

InLong Manager

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

flowers-59f:
Fixes https://github.com/apache/inlong/issues/12106

### Modifications
Before saving the connection, verify if there is a corresponding cluster
The reason for conducting the state judgment first is that some modes do not require the creation or deletion of MQ resources.
Refer to QueueResourceListener.listen(...)
### Verifying this change
I didn't write unit tests, but conducted a front-end and back-end integration test. 
The results are as follows:
Adding data access when there is no corresponding MQ cluster
image
image
image
Can intercept the creation operation

After creating the corresponding MQ cluster
image
image
Success in creation
And it will not affect the creation of data flow groups during data synchronization.
image

### Documentation

### DIFF
--- a/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
+++ b/inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
@@ -215,7 +215,27 @@ public class InlongGroupServiceImpl implements InlongGroupService {
         if (request.getEnableZookeeper() == null) {
             request.setEnableZookeeper(enableZookeeper ? InlongConstants.ENABLE_ZK : InlongConstants.DISABLE_ZK);
         }
-        InlongGroupOperator instance = groupOperatorFactory.getInstance(request.getMqType());
+
+        if (request.getEnableCreateResource() == null) {
+            request.setEnableCreateResource(InlongConstants.ENABLE_CREATE_RESOURCE);
+        }
+
+        if (request.getInlongGroupMode() == null) {
+            request.setInlongGroupMode(InlongConstants.STANDARD_MODE);
+        }
+
+        String mqType = request.getMqType();
+        if (InlongConstants.STANDARD_MODE.equals(request.getInlongGroupMode())
+                && InlongConstants.ENABLE_CREATE_RESOURCE.equals(request.getEnableCreateResource())) {
+            String clusterTag = request.getInlongClusterTag();
+            List<InlongClusterEntity> clusterEntities = clusterEntityMapper.selectByKey(clusterTag, null, mqType);
+            if (CollectionUtils.isEmpty(clusterEntities)) {
+                throw new BusinessException(String.format("cannot find any cluster by tag %s and type %s",
+                        clusterTag, mqType));
+            }
+        }
+
+        InlongGroupOperator instance = groupOperatorFactory.getInstance(mqType);
         groupId = instance.saveOpt(request, operator);
 
         // save ext info


inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
```diff
@@ -215,7 +215,27 @@ public String save(InlongGroupRequest request, String operator) {
...
             request.setEnableZookeeper(enableZookeeper ? InlongConstants.ENABLE_ZK : InlongConstants.DISABLE_ZK);
         }
-        InlongGroupOperator instance = groupOperatorFactory.getInstance(request.getMqType());
+
+        if (request.getEnableCreateResource() == null) {
+            request.setEnableCreateResource(InlongConstants.ENABLE_CREATE_RESOURCE);
+        }
+
+        if (request.getInlongGroupMode() == null) {
+            request.setInlongGroupMode(InlongConstants.STANDARD_MODE);
```
fuweng11, line 224: There is no need to set a default value here, as the create table statement for the inlong_group table has already set a default value

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
```diff
@@ -215,7 +215,27 @@ public String save(InlongGroupRequest request, String operator) {
...
+        }
+
+        String mqType = request.getMqType();
+        if (InlongConstants.STANDARD_MODE.equals(request.getInlongGroupMode())
```
fuweng11, line 228: When saving the group, no inlong_cluster_tag was assigned, so there is no need for judgment here.

inlong-manager/manager-service/src/main/java/org/apache/inlong/manager/service/group/InlongGroupServiceImpl.java
vernedeng: It is best to standardize the handling of all fields: either rely entirely on database default values, or configure the default values ​​within the application logic.

flowers-59f: In QueueResourceListener.listen()
```java
        if (InlongConstants.DATASYNC_REALTIME_MODE.equals(groupInfo.getInlongGroupMode())
                || InlongConstants.DATASYNC_OFFLINE_MODE.equals(groupInfo.getInlongGroupMode())) {
            log.warn("skip to execute QueueResourceListener as sync mode {} (1 for realtime sync, 2 for offline sync) "
                    + "for groupId={}", groupInfo.getInlongGroupMode(), groupId);
            if (GroupOperateType.INIT.equals(operateType)) {
                this.createQueueForStreams(groupInfo, groupProcessForm.getStreamInfos(), operator);
            }
            return ListenerResult.success("skip - disable create mq resource for sync mode");
        }

        if (InlongConstants.DISABLE_CREATE_RESOURCE.equals(groupInfo.getEnableCreateResource())) {
            log.warn("skip to execute QueueResourceListener as disable create resource for groupId={}", groupId);
            return ListenerResult.success("skip - disable create resource");
        }
```
In other modes, the creation/deletion of MQ resources will be skipped, so we only need to check for the existence of the corresponding MQ in specific modes. But, under this save link, the EnableCreateResource and InlongGroupMode passed in the request do not have values, so I set them according to the default values and made a check. Since there are other places that call this function, they might have set the corresponding values, and in those cases, validation might not be required.

This is my idea. Please take a look and see if there are any issues and let me know how I can improve.Thank you very much.

fuweng11: When saving the group, no clustertag is assigned, so this verification is unnecessary. Additionally, since the inlong_group table has default values set, there is no need to explicitly set them in the code during saving.
企业微信截图_ab049c8e-93eb-4d43-bcea-80d464e84832

flowers-59f: Yes. The database will assign a default value to this field when no value is set. However, it may also be assigned another value, so I think we still need to add a judgment here.But we can remove the code that sets the default value and adjust the judgment logic as follows.
```java
if ((request.getInlongGroupMode() == null || InlongConstants.STANDARD_MODE.equals(request.getInlongGroupMode()))  
&& (request.getEnableCreateResource() == null || InlongConstants.ENABLE_CREATE_RESOURCE.equals(request.getEnableCreateResource()))) {
    String clusterTag = request.getInlongClusterTag();
    List<InlongClusterEntity> clusterEntities = clusterEntityMapper.selectByKey(clusterTag, null, mqType);
    if (CollectionUtils.isEmpty(clusterEntities)) {
         throw new BusinessException(String.format("cannot find any cluster by tag %s and type %s",
         clusterTag, mqType));
    }
}
```
Here's my opinion.

Or, there's no need to set a default value and verify GroupMode and EnableCreateResource. Just proceed directly with the following checks.
```java
String clusterTag = request.getInlongClusterTag();
List<InlongClusterEntity> clusterEntities = clusterEntityMapper.selectByKey(clusterTag, null, mqType);
if (CollectionUtils.isEmpty(clusterEntities)) {
     throw new BusinessException(String.format("cannot find any cluster by tag %s and type %s",
     clusterTag, mqType));
}
```

